PR #3, opened 2024-10-18
Introduces pipeline.yml Github actions pipeline

markkasaboski:


markkasaboski:
I think we should go ahead and merge this in its broken state since this MR was about hooking up the pipeline and not fixing the errors. We can fix the errors after ;)

markkasaboski:
The publish.yml is a workflow that will get triggered when we create a new release from the Releases page. Right now, it will simply echo a message. We have to submit the first version of the app via the UI in Splunkbase. Once it gets approved, it will then have an AppID. We need the AppID in order to submit new releases via API.

ci/requirements.txt
TyMarc, line 1: All of these dependencies are needed for mypy/ruff? Won't they download their own subdependenceis?

PR #5, opened 2024-10-22
Added FlareSDK and replaced current feed retrieval implementation with the SDK

TyMarc:


flare_splunk_integration/appserver/static/javascript/views/configurationFileHelper.js
```diff
@@ -21,13 +21,13 @@ async function updateConfigurationFile(
   configurations = await promisify(configurations.fetch)();
```
markkasaboski, line 21: Just a note: I believe this `promisify()` outputs a message saying "DO NOT USE THIS IN PRODUCTION". We need to replace this at some point.

TyMarc: Definitely; I was thinking of refactoring the configuration screen next to polish those things out

markkasaboski (approved):

PR #6, opened 2024-10-22
Created vendor directory instead of symlinks to have same configuration between local and production

TyMarc:
Also makes mypy configuration easier since we can exclude the vendor folder. If you already have a local configuration, this command will clear the symlinks:
`find "flare_splunk_integration/bin" -type l -exec rm {} +`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,5 +6,4 @@ flare_splunk_integration/metadata/local.meta
 __pycache__/
 dependencies_temp
 
-flare_splunk_integration/bin/*
-!flare_splunk_integration/bin/*.py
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -6,5 +6,4 @@ flare_splunk_integration/metadata/local.meta
 __pycache__/
 dependencies_temp
 
-flare_splunk_integration/bin/*
-!flare_splunk_integration/bin/*.py
```
markkasaboski, line 10: Why are we removing this again?

markkasaboski (approved):
Nice!

PR #7, opened 2024-10-23
Fixes ruff linting errors

markkasaboski:


.github/workflows/pipeline.yml
```diff
@@ -46,5 +47,5 @@ jobs:
       - name: Splunk AppInspect
         run: make validate
 
-      - name: Test
-        run: make test
+      # - name: Test
+      #   run: CI=true make test
```
TyMarc: What's the CI=true part?

TyMarc (approved):

.github/workflows/pipeline.yml
markkasaboski: Screenshot 2024-10-24 at 10 08 40 AM

markkasaboski: Ah! I don't think that'll make it through Make, though. I'll probably have to create the var in the Makefile.

TyMarc: That should work yes!

markkasaboski: Ah! https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

The Github action runners have this as a default, BUT this doesn't mean that Make will provide it to the command, since every line in a makefile is executed in a subshell: https://web.mit.edu/gnu/doc/html/make_5.html#SEC44:~:text=When%20it%20is%20time%20to%20execute%20commands%20to%20update%20a%20target%2C%20they%20are%20executed%20by%20making%20a%20new%20subshell%20for%20each%20line.%20(In%20practice%2C%20make%20may%20take%20shortcuts%20that%20do%20not%20affect%20the%20results.)

See this: https://unix.stackexchange.com/questions/392050/set-environment-variable-for-subshell

... I'm gonna take a slightly different approach on this

PR #8, opened 2024-10-23
Moved the scripting logic back into the Makefile and separated app venv from ci venv

TyMarc:
Also deleted the stubs and python packaging dependencies for the application


flare_splunk_integration/flare_splunk_integration
```diff
@@ -0,0 +1 @@
+/Users/marc-antoinehinse/repos/splunk-integration/flare_splunk_integration
```
markkasaboski: What's this?

requirements.tools.txt
```diff
@@ -1,3 +1,4 @@
+pytest==8.3.2
```
markkasaboski, line 1: Oh! Nice catch. Missed that.

flare_splunk_integration/flare_splunk_integration
TyMarc: Oh wow good catch

markkasaboski (approved):
Just that one file I'm not sure about. Otherwise, LGTM

PR #10, opened 2024-10-24
Created a two-step configuration wizard, modified navbar to change configuration

TyMarc:
Also modified the splunk-local target in Makefile to depend on the venv and disable caching locally. 
Created a Rest API handler to be able to call the Flare SDK endpoints from Javascript
Created a Home page with Lorem Ipsum


Makefile
```diff
@@ -25,6 +25,7 @@ clean:
 	@rm -rf venv
 	@rm -rf venv-tools
 	@rm -rf flare_splunk_integration/bin/vendor
+	@unlink "/Applications/Splunk/etc/apps/flare_splunk_integration" || true
```
markkasaboski, line 28: Ah! Good call.

flare_splunk_integration/appserver/static/javascript/views/setupConfiguration.js
```diff
@@ -56,7 +68,34 @@ async function saveConfiguration(splunkSdk, serverKey, tenantId) {
...
+      const passwordId = storageRealm + ":" + key + ":";
+      storage.del(passwordId);
+    }
+
+    while (passwordExists) {
+      storage = await promisify(storage.fetch)();
+      passwordExists = doesPasswordExist(storage, storageRealm, key);
+    }
+
     storage.create({
```
markkasaboski, line 118: We need some linting on the javascript. Let's format this a bit better.

flare_splunk_integration/appserver/static/javascript/views/setupConfiguration.js
```diff
@@ -56,7 +68,34 @@ async function saveConfiguration(splunkSdk, serverKey, tenantId) {
...
+    var passwordExists = doesPasswordExist(storage, storageRealm, key);
+    if (passwordExists) {
+      const passwordId = storageRealm + ":" + key + ":";
+      storage.del(passwordId);
+    }
+
+    while (passwordExists) {
+      storage = await promisify(storage.fetch)();
+      passwordExists = doesPasswordExist(storage, storageRealm, key);
+    }
```
markkasaboski: A few things here:
1 - Looks to be a mix of 4 and 2 spaces or maybe tab and 2 spaces. 
2 - We should probably add an `.editorconfig` file.
3 - I remember you explaining to me that you can't just update a value in the password storage, but this looks a bit off to me. Why the while loop? If it does need to work this way, it's probably worth a comment to explain to whoever else comes through here to maintain it.

mypy.ini
```diff
@@ -1,3 +1,4 @@
 [mypy]
 exclude = (vendor*)/$
 follow_imports = skip
+mypy_path = flare_splunk_integration/bin/vendor
```
markkasaboski, line 4: Why is this here? The mypy we're expecting to use should be in `venv-tools/bin`, shouldn't it?

flare_splunk_integration/appserver/static/javascript/views/setupConfiguration.js
markkasaboski: I'm assuming that `storage.delete()` is async and we can't await its response, so the while loop is us waiting until the delete has completed. Is that correct?

Would it be possible to `promisify(storage.del(passwordId))` like we do with `storage.fetch()`?

This makes me question all of the other storage functions now

mypy.ini
TyMarc: this configuration is to let it know where to find dependencies if they're not able to find them with the normal paths; that's how I fixed my issue of Mypy not finding the flareio dependency in the flare_external_requests.py imports

flare_splunk_integration/appserver/static/javascript/views/setupConfiguration.js
TyMarc: I added a comment explaining why this is done that way 👍

Makefile
TyMarc: Hehe I was tired of getting errors after calling the clean target

.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
...
+
+    // Insert spaces when pressing Tab. This setting is overriden
+    // based on the file contents when `editor.detectIndentation` is true.
+    "editor.insertSpaces": true,
+
+    // When opening a file, `editor.tabSize` and `editor.insertSpaces`
+    // will be detected based on the file contents. Set to false to keep
+    // the values you've explicitly set, above.
+    "editor.detectIndentation": false
+}
```
markkasaboski: This will be vscode specific. An actual `.editorconfig` is editor agnostic. They should all (most?) respect it. We're open sourcing this thing, so I think we should probably use `.editorconfig`

Ah. I see for VSCode, it's a plugin: https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig

markkasaboski (approved):
Just those couple of nits. Looks good otherwise.

mypy.ini
markkasaboski: Interesting.

PR #11, opened 2024-10-28
Fixes sourcetype to properly parse json data

markkasaboski:
- Removes `index=flare` from `inputs.conf` so now all data will be sent to the "main" index.
- Changes `source` to "Flare" so that events are easily searched for using `index="main" source="Flare"`
- Fixes sourcetype as it should be a "pretrained" Splunk type. Changed to `json_no_timestamp` and this now parses the json correctly and is now readable. 
Screenshot 2024-10-28 at 1 13 01 PM


TyMarc (approved):
LGTM

PR #14, opened 2024-10-29
Remodeled application architecture to use webpack to give ability to code in typescript and JSX React

TyMarc:
- Converted the static JSX-less configuration screen to React


.gitignore
```diff
@@ -2,9 +2,20 @@
 flare.tar.gz
 venv
 venv-tools
-flare/local
-flare/metadata/local.meta
+packages/flare/local
```
markkasaboski, line 5: packages? What's happening here?

Makefile
```diff
@@ -2,16 +2,20 @@
...
 venv: requirements.txt
 	python -m venv venv
 	venv/bin/pip install --upgrade pip
-	venv/bin/pip install --target flare/bin/vendor -r requirements.txt
-	@find flare/bin/vendor -type d -name "*.dist-info" -exec rm -r {} +
-	@find flare/bin/vendor -type d -name "__pycache__" -exec rm -r {} +
-	@rm -rf flare/bin/vendor/bin
-	@rm -rf flare/bin/vendor/packaging
-	@rm -rf flare/bin/vendor/*-stubs
+	venv/bin/pip install --target packages/flare/python/vendor -r requirements.txt
```
markkasaboski, line 13: Why the need to move into another folder? And why "packages"?

packages/configuration-screen/src/util.ts
```diff
@@ -0,0 +1,21 @@
...
+            }
+
+            args.push(callback);
+
+            fn.call(this, ...args);
+        });
+    };
+}
+
+export { promisify };
```
markkasaboski, line 21: Since we're now going to the extent of installing a bunch of other dependencies, what's one more? 😂 

https://www.npmjs.com/package/react-promise

TyMarc: The promisify method still makes sense in our case, it's to avoid having to listen to the callback on each request if we don't need to listen to the error. The react-promise seems more useful if methods are already Promises and we want to receive the error + know if the data is still loading

.gitignore
TyMarc: The Splunk NPX scaffolding uses lerna and their base folder for the packages is called packages

Makefile
```diff
@@ -72,30 +80,27 @@ test: venv-tools
...
+.PHONY: format setup-web
 format: venv-tools
 	venv-tools/bin/ruff check --fix --unsafe-fixes
 	venv-tools/bin/ruff format
+	yarn run format
 
 .PHONY: format-check
 format-check: venv-tools
 	venv-tools/bin/ruff check
 	venv-tools/bin/ruff format --check
```
markkasaboski, line 94: Does `yarn run format` have a flag for `check` that we can use in here so that the pipeline can check frontend formatting?

Makefile
```diff
@@ -72,30 +80,27 @@ test: venv-tools
...
 
 .PHONY: lint
-lint: mypy format-check
+lint: setup-web venv-tools mypy format-check
+	yarn run lint
```
markkasaboski, line 97: Ah this is probably it. Should maybe go into format-check

mypy.ini
```diff
@@ -1,4 +1,4 @@
 [mypy]
-exclude = (vendor*)/$
+exclude = (vendor*)/$|(flare/flare*)
```
markkasaboski: Can go within the first parenthesis like this: `(flare/flare|vendor*)/$`

Makefile
TyMarc: For CI,  the task was format:verify

packages/configuration-screen/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Change Log
+
+0.0.1 – Release date: TBA
+-------
+
+* Initial version
```
markkasaboski: Is this changelog necessary? I'm thinking we should have one changelog and one version to rule them all in root. What do you think?

packages/configuration-screen/README.md
```diff
@@ -0,0 +1,3 @@
+# ConfigurationScreen
+
+Add all information required to get started with @splunk/configuration-screen here.
```
markkasaboski: Same with READMEs. Also noteworthy: https://dev.splunk.com/enterprise/docs/releaseapps/splunkbase/optionsforsubmittingcontent/#Documentation-structure

packages/configuration-screen/src/ConfigurationScreen.jsx
```diff
@@ -0,0 +1,95 @@
+import React, { useEffect, useState } from 'react';
```
markkasaboski: Should these not be using typescript?

packages/configuration-screen/src/ConfigurationScreenStyles.js
```diff
@@ -0,0 +1,19 @@
+import styled from 'styled-components';
```
markkasaboski, line 1: Whoa! We have to put styling in JS files?! 😱

packages/configuration-screen/src/setupConfiguration.ts
```diff
@@ -0,0 +1,141 @@
...
+    );
+}
+
+async function reloadApp(splunkService: SplunkService, appName: string): Promise<void> {
+    var splunkApps = splunkService.apps();
+    await promisify(splunkApps.fetch)();
+
+    var currentApp = splunkApps.item(appName);
+    await promisify(currentApp.reload)();
+}
```
markkasaboski: Did you write this or did `@splunk/create` make it? Looks interesting. Looks like the splunkApps object maintains its own state of fetched apps? Like, we don't get a list from it and then filter that list for our app?

packages/configuration-screen/src/setupConfiguration.ts
```diff
@@ -0,0 +1,141 @@
...
+    return service;
+}
+
+function retrieveUserTenants(
+    serverKey: string,
+    successCallback: (userTenants: Array<Tenant>) => void,
+    errorCallback: (errorMessage: string) => void
+): void {
+    const service = createService(applicationNameSpace);
+    const data = { serverKey: serverKey };
```
markkasaboski: Can't remember what we said. Are we changing these in this MR or get this MR merged first?

packages/configuration-screen/src/setupConfiguration.ts
```diff
@@ -0,0 +1,141 @@
...
+    }
+    return '';
+}
+
+function doesPasswordExist(
+    storage: SplunkStoragePasswordAccessors,
+    storageRealm: string,
+    key: string
+) {
+    const passwordId = storageRealm + ':' + key + ':';
```
markkasaboski: String concatenation is so old school. Are you a ye or a ne on template literals? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

markkasaboski (approved):
LGTM just a few questions

packages/configuration-screen/src/ConfigurationScreenStyles.js
TyMarc: We don't have to! It just lets us create named elements with a style if we don't want to add more logic to it. Like the error div I've added

packages/configuration-screen/src/setupConfiguration.ts
TyMarc: It's code that comes from the example we used in the PoC, apparently the token that our SplunkService has gives us access to the list of installed apps. Not sure if we can modify objects in them though

packages/configuration-screen/src/setupConfiguration.ts
TyMarc: A big ye, will change it!

packages/configuration-screen/src/setupConfiguration.ts
TyMarc: Those changes will be part of this PR: https://github.com/Flared/splunk-integration/pull/16, I'm waiting for this PR to be done to push my local changes

packages/configuration-screen/src/ConfigurationScreen.jsx
TyMarc: You're right, will change it

PR #16, opened 2024-10-30
flare.py: add missing typing + splunk useragent

aviau:
In this PR:
- Add missing typing in FlareAPI's constructor.
- Add `flare-splunk` to the user agent so that we can monitor it.
- Remove `self.api_key` and `self.tenant_id`. What were they used for...?
- Raise an exception if we don't have an API key. Mypy should have warned about this but typing was missing so it never raised.

This PR is completely untested. So it would be helpful if one of you can test it. Or feel free to just integrate my suggested changes yourself in separate commits.


flare/bin/flare.py
```diff
@@ -11,27 +10,51 @@
...
+
 class FlareAPI(AuthBase):
-    def __init__(self, *, app):
+    def __init__(self, *, app: t.Any) -> None:
         # Should be able to use app.service.storage_passwords.get(),
         # but I can't seem to get that to work. list() works.
-        server_key: Optional[str] = None
-        tenant_id: Optional[str] = None
+        api_key: t.Optional[str] = None
+        tenant_id: t.Optional[int] = None
```
aviau: How did this even work? FlareApiClient requires an int for the tenant id, not a string.

flare/bin/flare.py
```diff
@@ -11,27 +10,51 @@
...
+                api_key = item.clear_password
 
             if item.content.username == "tenantid":
                 tenant_id = item.clear_password
 
+        if not api_key:
+            raise Exception("API key not found")
+
         self.flare_endpoints = app.service.confs["flare"]["endpoints"]
-        self.api_key = server_key
```
aviau: Why keep that variable? It is never used

flare/bin/flare.py
```diff
@@ -11,27 +10,51 @@
...
-        self.tenant_id = tenant_id
```
aviau: Why keep that varaible? it is never used

flare/bin/flare.py
```diff
@@ -11,27 +10,51 @@
...
+    def __init__(self, *, app: t.Any) -> None:
         # Should be able to use app.service.storage_passwords.get(),
         # but I can't seem to get that to work. list() works.
-        server_key: Optional[str] = None
-        tenant_id: Optional[str] = None
+        api_key: t.Optional[str] = None
+        tenant_id: t.Optional[int] = None
         for item in app.service.storage_passwords.list():
             if item.content.username == "serverkey":
-                server_key = item.clear_password
```
aviau: `server_key` is not very clear naming here. Let's call it api_key?

flare/bin/flare.py
```diff
@@ -11,27 +10,51 @@
...
+        or ""
+    )
+    if isinstance(current_user_agent, bytes):
+        current_user_agent = current_user_agent.decode()
+    api_client._session["User-Agent"] = f"{current_user_agent} flare-splunk"
+    return api_client
+
+
 class FlareAPI(AuthBase):
-    def __init__(self, *, app):
```
aviau: Missing typing here. Just use `Any` if you have no types.

flare/bin/flare.py
```diff
@@ -11,27 +10,55 @@
...
+    )
+    current_user_agent: t.Union[str, bytes] = (
+        api_client._session.headers.get(
+            "User-Agent",
+        )
+        or ""
+    )
+    if isinstance(current_user_agent, bytes):
+        current_user_agent = current_user_agent.decode()
+    api_client._session["User-Agent"] = f"{current_user_agent} flare-splunk"
```
aviau: It would be even better in a follow-up PR if we can use `flare-splunk/{version}`. Where `version` would be the  version of the app.

markkasaboski: The app is broken here. Seeing these messages

`TypeError: 'Session' object does not support item assignment`
`api_client._session["User-Agent"] = f"{current_user_agent} flare-splunk".strip()`

aviau: oops. this should be `api_client._session.headers["User-Agent"]`

flare/bin/flare.py
markkasaboski: If we're changing the variable, we should also change the storage name: `if item.content.username == "serverkey":` < these references to serverkey.

flare/bin/flare.py
markkasaboski: > This PR is completely untested. So it would be helpful if one of you can test it. Or feel free to just integrate my suggested changes yourself in separate commits.

Ah, just saw this. OK, will do. Thanks.

TyMarc:
That looks good to me; I have rebased locally your PR on my branch and renamed the variables to api_key instead of server_key

aviau:
Feel free to merge yourself, or even close my PR

flare/bin/flare.py
aviau: Yes we should. "serverkey" does not really make it clear this is an api key.

TyMarc (approved):
LGTM

PR #18, opened 2024-10-31
Event ingestion flow

TyMarc:
- The cronjob runs every minute, but the script will early exit if it's been less than 10 minutes since the last event fetch
- The start date for the 'time' parameter is reset to today whenever the tenant is changed
- The 'next' value is saved per tenant to avoid duplicate events when switching between tenants
- Retrieving 50 events per request to reduce the amount of requests done on the Flare backend

### DIFF
--- a/packages/flare/python/cron_job_ingest_events.py
+++ b/packages/flare/python/cron_job_ingest_events.py
@@ -1,23 +1,44 @@
 import json
-import sys
 import os
+import sys
+import time
+
+from datetime import date
+from datetime import datetime
+from datetime import timedelta
+from enum import Enum
+from typing import Any
 from typing import Optional
+
 
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "vendor"))
 import vendor.splunklib.client as client
 
 from flare import FlareAPI
 
+
 APP_NAME = "flare"
 HOST = "localhost"
 SPLUNK_PORT = 8089
 REALM = APP_NAME + "_realm"
-KV_COLLECTION_NAME = "feednext"
+KV_COLLECTION_NAME = "event_ingestion_collection"
+CRON_JOB_THRESHOLD_SINCE_LAST_FETCH = timedelta(minutes=10)
 
 
-def main():
+class PasswordKeys(Enum):
+    API_KEY = "api_key"
+    TENANT_ID = "tenant_id"
+
+
+class CollectionKeys(Enum):
+    CURRENT_TENANT_ID = "current_tenant_id"
+    START_DATE = "start_date"
+    NEXT_TOKEN = "next_"
+    TIMESTAMP_LAST_FETCH = "timestamp_last_fetch"
+
+
+def main() -> None:
     try:
-        # Example using a token
         splunk_service = client.connect(
             host=HOST,
             port=SPLUNK_PORT,
@@ -29,44 +50,163 @@ def main():
         raise Exception(str(e))
 
     app: client.Application = splunk_service.apps[APP_NAME]
-    flare_api = FlareAPI(app=app)
+    create_collection(app=app)
 
-    from_ = get_from_value(app=app)
-    event_feed = flare_api.retrieve_feed(from_=from_)
-    set_from_value(app=app, next=event_feed["next"])
+    # To avoid cron jobs from doing the same work at the same time, exit new cron jobs if a cron job is already doing work
+    last_fetched_timestamp = get_last_fetched(app)
+    if last_fetched_timestamp and last_fetched_timestamp < (
+        datetime.now() - CRON_JOB_THRESHOLD_SINCE_LAST_FETCH
+    ):
+        return
 
-    if event_feed["items"]:
-        for item in event_feed["items"]:
-            print(json.dumps(item))
+    api_key: Optional[str] = None
+    tenant_id: Optional[int] = None
+    for item in app.service.storage_passwords.list():
+        if item.content.username == PasswordKeys.API_KEY.value:
+            api_key = item.clear_password
+
+        if item.content.username == PasswordKeys.TENANT_ID.value:
+            tenant_id = (
+                int(item.clear_password) if item.clear_password is not None else None
+            )
+
+    if not api_key:
+        raise Exception("API key not found")
+
+    if not tenant_id:
+        raise Exception("Tenant ID not found")
+
+    try:
+        flare_api = FlareAPI(app=app, api_key=api_key, tenant_id=tenant_id)
+
+        next = get_next(app=app, tenant_id=tenant_id)
+        start_date = get_start_date(app=app)
+        for response in flare_api.retrieve_feed(next=next, start_date=start_date):
+            save_last_fetched(app=app)
+
+            # Rate limiting.
+            time.sleep(1)
+
+            if response.status_code != 200:
+                print(response.text, file=sys.stderr)
+                return
+
+            event_feed = response.json()
+            save_start_date(app=app, tenant_id=tenant_id)
+            save_next(app=app, tenant_id=tenant_id, next=event_feed["next"])
+
+            if event_feed["items"]:
+                for item in event_feed["items"]:
+                    print(json.dumps(item))
+    except Exception as e:
+        print("Exception={}".format(e))
 
 
-def get_from_value(app: client.Application) -> Optional[str]:
-    from_: Optional[str] = None
-    if KV_COLLECTION_NAME in app.service.kvstore:
-        data = app.service.kvstore[KV_COLLECTION_NAME].data.query()
-        if len(data) > 1:
-            from_ = data[0]["value"]
-        else:
-            app.service.kvstore.delete(KV_COLLECTION_NAME)
-
-    return from_
+def get_next(app: client.Application, tenant_id: int) -> Optional[str]:
+    return get_collection_value(
+        app=app, key="{}{}".format(CollectionKeys.NEXT_TOKEN.value, tenant_id)
+    )
 
 
-def set_from_value(app: client.Application, next: Optional[str]):
+def get_start_date(app: client.Application) -> Optional[str]:
+    return get_collection_value(app=app, key=CollectionKeys.START_DATE.value)
+
+
+def get_current_tenant_id(app: client.Application) -> Optional[int]:
+    current_tenant_id = get_collection_value(
+        app=app, key=CollectionKeys.CURRENT_TENANT_ID.value
+    )
+    try:
+        return int(current_tenant_id) if current_tenant_id else None
+    except Exception:
+        pass
+    return None
+
+
+def get_last_fetched(app: client.Application) -> Optional[datetime]:
+    timestamp_last_fetched = get_collection_value(
+        app=app, key=CollectionKeys.TIMESTAMP_LAST_FETCH.value
+    )
+    if timestamp_last_fetched:
+        try:
+            return datetime.fromisoformat(timestamp_last_fetched)
+        except Exception:
+            pass
+    return None
+
+
+def create_collection(app: client.Application) -> None:
     if KV_COLLECTION_NAME not in app.service.kvstore:
         # Create the collection
         app.service.kvstore.create(
             name=KV_COLLECTION_NAME, fields={"_key": "string", "value": "string"}
         )
-        # Insert
-        app.service.kvstore[KV_COLLECTION_NAME].data.insert(
-            json.dumps({"_key": "next", "value": next})
+
+
+def save_start_date(app: client.Application, tenant_id: int) -> None:
+    current_tenant_id = get_current_tenant_id(app=app)
+    # If this is the first request ever, insert today's date so that future requests will be based on that
+    if not get_start_date(app):
+        save_collection_value(
+            app=app,
+            key=CollectionKeys.START_DATE.value,
+            value=datetime.today().isoformat(),
         )
-    elif not next:
-        app.service.kvstore[KV_COLLECTION_NAME].data.delete(id="next")
+
+    # If the current tenant has changed, update the start date so that future requests will be based off today
+    # If you switch tenants, this will avoid the old tenant from ingesting all the events before today and the day
+    # that tenant was switched in the first place.
+    if current_tenant_id != tenant_id:
+        app.service.kvstore[KV_COLLECTION_NAME].data.update(
+            id=CollectionKeys.START_DATE.value,
+            data=json.dumps({"value": date.today().isoformat()}),
+        )
+
+
+def save_next(app: client.Application, tenant_id: int, next: Optional[str]) -> None:
+    # If we have a new next value, update the collection for that tenant to continue searching from that point
+    if not next:
+        return
+
+    save_collection_value(
+        app=app,
+        key="{}{}".format(CollectionKeys.NEXT_TOKEN.value, tenant_id),
+        value=next,
+    )
+
+
+def save_last_fetched(app: client.Application) -> None:
+    save_collection_value(
+        app=app,
+        key=CollectionKeys.TIMESTAMP_LAST_FETCH.value,
+        value=datetime.now().isoformat(),
+    )
+
+
+def get_collection_value(app: client.Application, key: str) -> Optional[str]:
+    if KV_COLLECTION_NAME in app.service.kvstore:
+        data = app.service.kvstore[KV_COLLECTION_NAME].data.query()
+        for entry in data:
+            if entry["_key"] == key:
+                return entry["value"]
+
+    return None
+
+
+def save_collection_value(app: client.Application, key: str, value: Any) -> None:
+    if not get_collection_value(app=app, key=key):
+        app.service.kvstore[KV_COLLECTION_NAME].data.insert(
+            json.dumps(
+                {
+                    "_key": key,
+                    "value": value,
+                }
+            )
+        )
     else:
         app.service.kvstore[KV_COLLECTION_NAME].data.update(
-            id="next", data=json.dumps({"value": next})
+            id=key,
+            data=json.dumps({"value": value}),
         )
 
 
--- a/packages/flare/python/flare.py
+++ b/packages/flare/python/flare.py
@@ -1,12 +1,11 @@
-from http.client import HTTPMessage
+import requests
 import typing as t
-import sys
+import vendor.splunklib.client as client
 
+from datetime import date
+from vendor.flareio import FlareApiClient
 from vendor.requests.auth import AuthBase
 
-from urllib.error import HTTPError
-from vendor.flareio import FlareApiClient
-import vendor.splunklib.client as client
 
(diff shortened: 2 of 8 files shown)


packages/flare/python/cron_job_ingest_events.py
```diff
@@ -2,6 +2,7 @@
 import sys
 import os
 from typing import Optional
+from datetime import date, datetime
```
markkasaboski: Hmmm, I think we generally prefer to have one import per line. We do on Pyro anyway. We may need to add that check to MyPy or Ruff.

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +32,119 @@ def main():
...
-            print(json.dumps(item))
+    if not api_key:
+        raise Exception("API key not found")
 
+    if not tenant_id:
+        raise Exception("Tenant ID not found")
+    flare_api = FlareAPI(app=app, api_key=api_key, tenant_id=tenant_id)
 
-def get_from_value(app: client.Application) -> Optional[str]:
+    from_ = get_from_value(app=app, tenant_id=tenant_id)
```
markkasaboski: I won't make too much of a fuss if you want to keep it this way, but two things:

1. I would prefer to see "next" instead of "from". But i'll leave that up to you.
2. Can it just be `get_next()` or `get_from()`? 

Same for other functions below, but I'll get to those.

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +32,119 @@ def main():
...
+    start_date = get_start_date_value(app=app)
```
markkasaboski: `get_start_date()`?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +32,119 @@ def main():
...
+    while True:
+        event_feed = flare_api.retrieve_feed(from_=from_, start_date=start_date)
+        save_values(app=app, tenant_id=tenant_id, next=event_feed["next"])
```
markkasaboski: `save_state()`? 

At first glance, I'm thinking that this function is doing too many things. I'd be tempted to break it up into several other functions. If it were a case for atomicity, one function and a bunch of logic to ensure atomicity would make sense, but that's not what's happening here. 

What do you think?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +32,119 @@ def main():
...
+            >= CRON_JOB_MAX_RUNNING_SECONDS
+        ):
+            break
+        if event_feed["next"]:
+            from_ = event_feed["next"]
+        else:
+            break
+
+
+def get_from_value(app: client.Application, tenant_id: int) -> Optional[str]:
```
markkasaboski: `get_(next|from)()`?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +32,119 @@ def main():
...
+        for entry in data:
+            if entry["_key"] == "next_{}".format(tenant_id):
+                from_ = entry["value"]
+                break
 
     return from_
 
 
-def set_from_value(app: client.Application, next: Optional[str]):
+def get_start_date_value(app: client.Application) -> Optional[str]:
```
markkasaboski: `get_start_date()`?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +32,119 @@ def main():
...
+        data = app.service.kvstore[KV_COLLECTION_NAME].data.query()
+        for entry in data:
+            if entry["_key"] == "start_date":
+                start_date = entry["value"]
+                break
+
+    return start_date
+
+
+def get_current_tenant_id_value(app: client.Application) -> Optional[int]:
```
markkasaboski: `get_current_tenant_id()`?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +32,119 @@ def main():
...
+        data = app.service.kvstore[KV_COLLECTION_NAME].data.query()
+        for entry in data:
+            if entry["_key"] == "current_tenant_id":
+                current_tenant_id = int(entry["value"]) if entry["value"] else -1
+                break
+
+    return current_tenant_id
+
+
+def save_values(app: client.Application, tenant_id: int, next: Optional[str]):
```
markkasaboski: As stated above, I feel like this function is doing too much.

markkasaboski (approved):
The logic looks good me. Mostly just coding preferences. I'll leave it up to you if you wan to leave it as is.

packages/flare/python/cron_job_ingest_events.py
TyMarc: I split the methods in 3 separate methods instead, one for creating the collection, one to save the start date and the other to save the new next

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,23 +1,25 @@
...
 
 APP_NAME = "flare"
 HOST = "localhost"
 SPLUNK_PORT = 8089
 REALM = APP_NAME + "_realm"
 KV_COLLECTION_NAME = "feednext"
+CRON_JOB_MAX_RUNNING_SECONDS = 50
 
 
 def main():
```
aviau: Missing typing of the return value here

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,23 +1,25 @@
...
+    # Cron job are stacking, so you have to make sure to stop the current one before the next one starts
```
aviau: That looks really inefficient. Can we work around this in a smarter way? How about using the kvstore with a timestamp?
- When the job start, set a timestamp to the current date.
- After each request response, re-set the timestamp to the current date.
- When the cronjob start, look in the kv service, if the timestamp is less than 30 minutes ago, just exit right away.

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +31,125 @@ def main():
...
+            break
+        if event_feed["next"]:
+            next = event_feed["next"]
+        else:
+            break
 
-def get_from_value(app: client.Application) -> Optional[str]:
-    from_: Optional[str] = None
+
+def get_next(app: client.Application, tenant_id: int) -> Optional[str]:
```
aviau: It looks like you are copying the kvstore.data.query() pattern 3 times in this file. Can you extract it in a function?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +31,125 @@ def main():
...
+            if entry["_key"] == "start_date":
+                start_date = entry["value"]
+                break
+
+    return start_date
 
-def set_from_value(app: client.Application, next: Optional[str]):
+
+def get_current_tenant_id(app: client.Application) -> Optional[int]:
+    current_tenant_id: Optional[int] = -1
```
aviau: Why `-1` ? Python has `Optional` which supports `None`.

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +31,125 @@ def main():
...
+def create_collection(app: client.Application):
     if KV_COLLECTION_NAME not in app.service.kvstore:
         # Create the collection
         app.service.kvstore.create(
             name=KV_COLLECTION_NAME, fields={"_key": "string", "value": "string"}
         )
-        # Insert
+
+
+def save_start_date(app: client.Application, tenant_id: int):
```
aviau: missing return value typing here

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +31,125 @@ def main():
...
+        data = app.service.kvstore[KV_COLLECTION_NAME].data.query()
+        for entry in data:
+            if entry["_key"] == "current_tenant_id":
+                current_tenant_id = int(entry["value"]) if entry["value"] else -1
+                break
+
+    return current_tenant_id
+
+
+def create_collection(app: client.Application):
```
aviau: missing return value typing here

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,46 +31,125 @@ def main():
...
+    # If you switch tenants, this will avoid the old tenant from ingesting all the events before today and the day
+    # that tenant was switched in the first place.
+    if current_tenant_id != tenant_id:
         app.service.kvstore[KV_COLLECTION_NAME].data.update(
-            id="next", data=json.dumps({"value": next})
+            id="start_date", data=json.dumps({"value": date.today().isoformat()})
         )
 
 
+def save_next(app: client.Application, tenant_id: int, next: Optional[str]):
```
aviau: missing return value typing here

packages/flare/python/cron_job_ingest_events.py
TyMarc: I added configuration for ruff + isort dependency with its configuration for single line imports

packages/flare/python/cron_job_ingest_events.py
TyMarc: Instead of checking for a specific threshold to leave a cronjob, we could just raise a flag in the collection to say that we're actually doing work. If a new job spawns and the flag is raised, exit right away. WDYT?

aviau: That would break forever in some crash scenarios.

packages/flare/python/flare.py
```diff
@@ -36,41 +37,30 @@ def get_flare_api_client(
...
         url = self.flare_endpoints["me_feed_endpoint"]
-        response = self.flare_client.post(
+        params: t.Dict[str, t.Any] = {"lite": True, "size": 50}
+        if next:
+            params["search_after"] = next
+        else:
+            params["time"] = "{}@".format(
+                start_date if start_date else date.today().isoformat()
+            )
+        response = self.flare_client.get(
```
aviau: We should use .scroll() here.

See the docs:
- https://api.docs.flare.io/guides/global-search#python-sdk-example

this method could return an iterator of `cursor, results`

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,44 +33,133 @@ def main():
...
+    # that tenant was switched in the first place.
+    if current_tenant_id != tenant_id:
+        app.service.kvstore[KV_COLLECTION_NAME].data.update(
+            id="start_date", data=json.dumps({"value": date.today().isoformat()})
+        )
+
+
+def save_next(app: client.Application, tenant_id: int, next: Optional[str]) -> None:
+    # If we have a new next value, update the collection for that tenant to continue searching from that point
+    if next:
```
aviau: Instead of indenting the whole function, we usually reverse the check and return early.

```python
if not next:
    return
rest of code...
```

aviau (approved):

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,44 +33,133 @@ def main():
...
+            app.service.kvstore[KV_COLLECTION_NAME].data.insert(
+                json.dumps({"_key": "next_{}".format(tenant_id), "value": next})
+            )
+        else:
+            app.service.kvstore[KV_COLLECTION_NAME].data.update(
+                id="next_{}".format(tenant_id), data=json.dumps({"value": next})
+            )
+
+
+def set_job_running(app: client.Application, is_running: bool) -> None:
```
aviau: I don't agree with this design. This will be brittle in some crash scenarios.

packages/flare/python/flare.py
TyMarc: Per the documentation:
```
# Rate limiting.
time.sleep(1)
```

Is this necessary?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,44 +33,133 @@ def main():
...
+    create_collection(app=app)
+
+    # To avoid cron jobs from doing the same work at the same time, exit new cron jobs if a cron job is already doing work
+    if get_job_running(app):
+        return
+
+    api_key: Optional[str] = None
+    tenant_id: Optional[int] = None
+    for item in app.service.storage_passwords.list():
+        if item.content.username == "api_key":
```
aviau: Can we extract the magic strings to constants?

Maybe create an enum for the setting keys.

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,44 +33,133 @@ def main():
...
 
-    return from_
 
+def get_start_date(app: client.Application) -> Optional[str]:
+    return get_collection_value(app=app, key="start_date")
 
-def set_from_value(app: client.Application, next: Optional[str]):
+
+def get_current_tenant_id(app: client.Application) -> Optional[int]:
+    current_tenant_id = get_collection_value(app=app, key="current_tenant_id")
```
aviau: Let's extract the strings here to constants.

packages/flare/python/flare.py
aviau: For now yes

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,23 +1,27 @@
...
 import vendor.splunklib.client as client
 
 from flare import FlareAPI
 
+
 APP_NAME = "flare"
 HOST = "localhost"
 SPLUNK_PORT = 8089
 REALM = APP_NAME + "_realm"
 KV_COLLECTION_NAME = "feednext"
```
aviau: May I ask what this name is? did we select this name?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -29,44 +33,133 @@ def main():
...
+                int(item.clear_password) if item.clear_password is not None else None
+            )
+
+    if not api_key:
+        raise Exception("API key not found")
 
-    from_ = get_from_value(app=app)
-    event_feed = flare_api.retrieve_feed(from_=from_)
-    set_from_value(app=app, next=event_feed["next"])
+    if not tenant_id:
```
aviau, line 76: Do we really want to crash in this case? Should it be fine if the user wants to use his default tenant?

TyMarc: You cannot finish the configuration if you don't select a tenant. The UI dropdown will auto select the default tenant. The tenant ID is used to keep track of the 'next' token in case the user switches tenants

packages/flare/python/cron_job_ingest_events.py
TyMarc: We did during the PoC, will change it to something more obvious

packages/flare/python/cron_job_ingest_events.py
aviau: ok sure. Personally I would have pushed that to the nice-to-haves.

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,72 +1,207 @@
...
 
-def set_from_value(app: client.Application, next: Optional[str]):
+def get_current_tenant_id(app: client.Application) -> Optional[int]:
+    current_tenant_id = get_collection_value(
+        app=app, key=CollectionKeys.CURRENT_TENANT_ID.value
+    )
+    return int(current_tenant_id) if current_tenant_id else None
+
+
+def get_last_fetched(app: client.Application) -> Optional[int]:
```
aviau: Ideally this should return a `datetime.datetime`

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,72 +1,207 @@
...
     app: client.Application = splunk_service.apps[APP_NAME]
-    flare_api = FlareAPI(app=app)
+    create_collection(app=app)
 
-    from_ = get_from_value(app=app)
-    event_feed = flare_api.retrieve_feed(from_=from_)
-    set_from_value(app=app, next=event_feed["next"])
+    # To avoid cron jobs from doing the same work at the same time, exit new cron jobs if a cron job is already doing work
+    last_fetched_timestamp = get_last_fetched(app)
+    if (
```
aviau:
```python

CRON_JOB_THRESHOLD_SINCE_LAST_FETCH = timedelta(minutes=15)

if (
    last_fetched_timestamp
    and last_fetched_timestamp < (datetime.datetime.now() - timedelta(minutes=15)
):
    pass
```

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,72 +1,207 @@
...
+        app=app, key=CollectionKeys.CURRENT_TENANT_ID.value
+    )
+    return int(current_tenant_id) if current_tenant_id else None
+
+
+def get_last_fetched(app: client.Application) -> Optional[int]:
+    timestamp_last_fetched = get_collection_value(
+        app=app, key=CollectionKeys.TIMESTAMP_LAST_FETCH.value
+    )
+    return int(timestamp_last_fetched) if timestamp_last_fetched else None
```
aviau: We need to handle bad data here. Let's maybe wrap this in a try-except just to make sure the app still works if there is a bad value in get_collection_value

packages/flare/python/cron_job_ingest_events.py
markkasaboski: pass? 

Maybe I'm missing something, but wether this condition evaluates truthy or falsy, everything after it is still gonna run.

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,72 +1,207 @@
...
-    event_feed = flare_api.retrieve_feed(from_=from_)
-    set_from_value(app=app, next=event_feed["next"])
+    # To avoid cron jobs from doing the same work at the same time, exit new cron jobs if a cron job is already doing work
+    last_fetched_timestamp = get_last_fetched(app)
+    if (
+        last_fetched_timestamp
+        and int(datetime.now().timestamp()) - last_fetched_timestamp
+        < CRON_JOB_THRESHOLD_SINCE_LAST_FETCH
+    ):
+        pass
```
markkasaboski: pass?

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -1,72 +1,207 @@
...
-        # Example using a token
         splunk_service = client.connect(
             host=HOST,
             port=SPLUNK_PORT,
             app=APP_NAME,
             token=sys.stdin.readline().strip(),
         )
     except Exception as e:
+        # Use logger
         print(str(e), file=sys.stderr)
```
markkasaboski: Yeah, we can remove these.

packages/flare/python/cron_job_ingest_events.py
TyMarc: Yep good catch, I had switch my return to pass to test my code

PR #19, opened 2024-11-01
Created logger and constants file to avoid duplicating the APP_NAME variable everywhere

TyMarc:


markkasaboski (approved):
Awesome! LGTM

packages/flare/python/logger.py
```diff
@@ -0,0 +1,57 @@
...
+import tempfile
+
+from constants import APP_NAME
+from logging.handlers import TimedRotatingFileHandler
+from typing import Any
+
+
+class Logger:
+    def __init__(self, *, class_name: str) -> None:
+        SPLUNK_HOME = os.environ.get("SPLUNK_HOME")
```
aviau, line 12: This isn't a constant so maybe don't use allcaps here?

PR #20, opened 2024-11-01
Early exit condition was reversed + added debug logs

TyMarc:


packages/flare/python/cron_job_ingest_events.py
```diff
@@ -73,6 +73,12 @@ def main() -> None:
 
         next = get_next(app=app, tenant_id=tenant_id)
         start_date = get_start_date(app=app)
+        logger.debug(
```
aviau: I see a lot of `.format()` in your code. Can you remove them and instead use f-strings?

The advantage with f-strings is that mypy can validate them.

TyMarc: Didn't know about that one, it's way easier to read too

packages/flare/python/cron_job_ingest_events.py
```diff
@@ -73,6 +71,10 @@ def main() -> None:
...
+            f"Fetching Tenant:{tenant_id}, Next:{next}, Start Date:{start_date}"
```
aviau: Here you can just use `f"Fetching {tenant_id=} {next=} {start_date=}"` and it will automatically display with what you want.

aviau: docs: https://docs.python.org/3/whatsnew/3.8.html#f-strings-support-for-self-documenting-expressions-and-debugging

TyMarc: Nice 👍

aviau (approved):

PR #21, opened 2024-11-05
Introduces tests for cron_job_ingest_events.py plus refactoring for e…

markkasaboski:
…asier tests

When reviewing, please keep in mind that I had a good read over https://flareio.atlassian.net/wiki/spaces/engineering/pages/103874561/How+to+write+great+tests 

and followed some (not all) of its advice.

### DIFF
--- a/packages/flare/tests/python/test_ingest_events.py
+++ b/packages/flare/tests/python/test_ingest_events.py
@@ -1,0 +1,309 @@
+import json
+import os
+import pytest
+import sys
+
+from datetime import date
+from datetime import datetime
+from datetime import timedelta
+from typing import Any
+from unittest.mock import MagicMock
+from unittest.mock import Mock
+from unittest.mock import PropertyMock
+from unittest.mock import patch
+
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), "../../python"))
+from constants import CRON_JOB_THRESHOLD_SINCE_LAST_FETCH
+from constants import KV_COLLECTION_NAME
+from constants import CollectionKeys
+from cron_job_ingest_events import fetch_and_print_feed_results
+from cron_job_ingest_events import get_api_credentials
+from cron_job_ingest_events import get_collection_value
+from cron_job_ingest_events import get_current_tenant_id
+from cron_job_ingest_events import get_last_fetched
+from cron_job_ingest_events import get_start_date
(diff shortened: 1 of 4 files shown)


packages/flare/tests/python/test_ingest_events.py
```diff
@@ -0,0 +1,193 @@
...
+def test_fetch_and_print_feed_results_expect_exception() -> None:
+    logger = MagicMock()
+    app = MagicMock()
+    with patch.object(
+        app.service,
+        "confs",
+        {
+            "flare": {
+                "endpoints": {
+                    "me_feed_endpoint": "https://api.flare.io/firework/api2/me/feed"
```
TyMarc: this is not the value in that stanza

packages/flare/tests/python/test_ingest_events.py
```diff
@@ -0,0 +1,193 @@
...
+        flare_api_mock_instance.retrieve_feed.return_value = [response_mock]
+
+        with patch("time.sleep", return_value=None):
+            with patch.object(
+                app.service,
+                "confs",
+                {
+                    "flare": {
+                        "endpoints": {
+                            "me_feed_endpoint": "https://api.flare.io/firework/api2/me/feed"
```
TyMarc: this is not the value in that stanza

packages/flare/tests/python/test_ingest_events.py
```diff
@@ -0,0 +1,193 @@
...
```
TyMarc: this is not the value in that stanza

TyMarc: actually I'm not entirely sure what we're testing in those 3 tests because we force the expected response from the beginning

markkasaboski: You have to mock a response from a request since it's not ideal to use or get real data as this would require setting up a specific test account on the app where we can ensure that the api key and tenant will never be deleted. This is not something you'd ever want to do in a unit test. So this is just exercising the code following the response validating that is working as we expect it to.

TyMarc (approved):
LGTM, just need to reformat the file

TyMarc (changes requested):
We should test what happens if we put garbage in the start_date key and try to retrieve it. It shouldn't crash and return None

TyMarc (approved):

PR #25, opened 2024-11-08
Refactored some of the python architecture

TyMarc:
- Removed endpoint config (flare)
- Using urllib to parse query parameters (flare_external_requests)
- Renamed local variables to removed all caps (flare_external_requests)
- Since endpoint config is removed, the tenants requests is done inside flare.py
- Created static method to generate get_next_token() collection key (constants)
- Moved the handling of the event_feed in the main method of the cron job (cron_job_ingest_events)
- Calling save_last_fetched before the first fetch to avoid race conditions (cron_job_ingest_events)
- Always pass the filter params in the fetching of the feed to make sure that the 'next' token is valid (flare)
- Removed our status code check and handling the exception from raise_for_status (cron_job_ingest_events)

### DIFF
--- a/packages/flare/bin/cron_job_ingest_events.py
+++ b/packages/flare/bin/cron_job_ingest_events.py
@@ -6,6 +6,7 @@ import time
 from datetime import date
 from datetime import datetime
 from typing import Any
+from typing import Iterator
 from typing import Optional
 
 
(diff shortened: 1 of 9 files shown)


packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -32,16 +33,33 @@ def main(logger: Logger, app: client.Application) -> None:
...
+
+        # Rate limiting.
+        time.sleep(1)
+
+        save_start_date(app=app, tenant_id=tenant_id)
+        save_next(app=app, tenant_id=tenant_id, next=event_feed["next"])
+
+        if event_feed["items"]:
+            for item in event_feed["items"]:
+                print(json.dumps(item))
```
aviau: I am wondering if flushing here  would help with your "bug" of failing import during an exception.

You can `print(item, flush=True)`

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -179,43 +197,25 @@ def save_collection_value(app: client.Application, key: str, value: Any) -> None
...
 
 
-def fetch_and_print_feed_results(
+def fetch_feed(
     logger: Logger,
     app: client.Application,
     api_key: str,
     tenant_id: int,
-) -> None:
+) -> Iterator[Any]:
```
aviau: Let's say Iterator[dict]

aviau (approved):

PR #26, opened 2024-11-08
Switched to v4 feed fetching and toggle for metadata only

TyMarc:


TyMarc:
Missing Unit testing for the flare.py class since we're now doing more work than just a single request

packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -0,0 +1,135 @@
...
+@patch("flare.FlareAPI._retrieve_event_feed_metadata")
+def test_flare_full_data_without_metadata(
+    retrieve_event_feed_metadata_mock: MagicMock,
+    enrich_event_from_metadata_uid_mock: MagicMock,
+) -> None:
+    expected_return_value = {
+        "next": "some_next_value",
+        "items": [
+            {
+                "bad_data": "foo",
```
markkasaboski, line 25: What is this "bad_data"?

TyMarc: it's a payload that doesn't have the [metadata][uid] in it basically

markkasaboski (approved):
LGTM

packages/flare/bin/flare.py
```diff
@@ -40,23 +42,73 @@ def __init__(self, *, api_key: str, tenant_id: Optional[int] = None) -> None:
...
+        ingest_metadata_only: bool,
+    ) -> Iterator[dict]:
+        for response in self._retrieve_event_feed_metadata(
+            next=next,
+            start_date=start_date,
+        ):
+            event_feed = response.json()
+            self.logger.debug(event_feed)
+            if ingest_metadata_only:
+                yield event_feed
```
aviau, line 61: Should we yield items instead?

```python
for item in event_feed["items"]:
    yield item
```


Right now your function returns one of:
- a full page
- an item

Intead it should return one of:
- A full item
- A partial item

TyMarc: i thought about it but I preferred to keep it that way so that it's easier to track the 'next' token and avoid constantly saving the last_fetched and start_date, but mostly for the next token

packages/flare/bin/flare.py
```diff
@@ -40,23 +42,73 @@ def __init__(self, *, api_key: str, tenant_id: Optional[int] = None) -> None:
...
+        for response in self.flare_client.scroll(
+            method="POST",
+            url="/firework/v4/events/tenant/_search",
+            json=data,
+        ):
+            yield response
+            # Rate limiting.
+            time.sleep(1)
+
+    def _enrich_event_from_metadata_uid(self, *, event: dict) -> dict:
```
aviau, line 102: I don't think we want to enrich. We just want to return the full item.

packages/flare/tests/bin/test_flare_wrapper.py
markkasaboski: As a developer coming in here to maintain this code, would/should it be obvious to me why this is considered "bad_data"?

The test accepts the fact that this is bad data, but I don't know why.

PR #32, opened 2024-11-12
Creates the flare index

markkasaboski:


markkasaboski:
@aviau I'm gonna get to fixing the unit tests and refactoring to use Protocol. #comingsoon

packages/configuration-screen/src/components/ConfigurationCompletedStep.tsx
```diff
@@ -10,6 +10,17 @@ const ConfigurationCompletedStep: FC<{
     tenantName: string;
     onBackClicked: () => void;
 }> = ({ show, tenantName, onBackClicked }) => {
+    let [index, setIndex] = useState<string>();
```
TyMarc: nit: can we call it indexName? Otherwise it sounds like an index for a loop

packages/configuration-screen/src/components/ConfigurationCompletedStep.tsx
```diff
@@ -10,6 +10,17 @@ const ConfigurationCompletedStep: FC<{
...
+
+    useEffect(() => {
+        const getIndex = async () => {
```
TyMarc: getFlareDataUrl().then((indexName) => setIndexName(indexName));

packages/configuration-screen/src/utils/configurationFileHelper.ts
```diff
@@ -138,3 +138,29 @@ export async function updateConfigurationFile(
     // change the information of a property
     await updateStanzaProperties(configurationStanzaAccessor, properties);
 }
+
+export async function getCurrentIndex(service: SplunkService): Promise<string> {
```
TyMarc: nit: getCurrentIndexName

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -62,6 +63,11 @@ def main(logger: Logger, app: client.Application) -> None:
     logger.info(f"Retrieved {events_retrieved_count} events")
 
 
+def create_flare_index(service: Service) -> None:
+    if APP_NAME not in service.indexes:
```
TyMarc, line 67: we should probably make sure we only create the index once since they might decide to use their own index and delete ours. Could we add a flag in the Collection to say that we did it once?

TyMarc (changes requested):

TyMarc (approved):
LGTM

PR #35, opened 2024-11-13
Created the status page, showing basic information of when was the last time the events were ingested

TyMarc:
Normal view:
image

Advanced view:
image


packages/flare/package.json
```diff
@@ -15,6 +15,7 @@
   "devDependencies": {
     "@babel/core": "^7.2.0",
     "@splunk/configuration-screen": "^0.0.1",
+    "@splunk/status-screen": "^0.0.1",
```
markkasaboski: I just noticed this... Our screens are different packages?

Two questions:

Should these be in devDependencies?
Should these be under an @flare/ namespace?

markkasaboski: I'm watching a Lerna vid right now cuz I know nothing about it.

packages/status-screen/src/StatusScreen.tsx
```diff
@@ -0,0 +1,87 @@
...
+                                );
+                                return parsedTenantId === id;
+                            }
+                            return true;
+                        })
+                        .reverse()
+                );
+            });
+        });
+    }, []);
```
markkasaboski, line 46: https://www.reddit.com/r/ProgrammerHumor/comments/27yykv/indent_hadouken/ 😂 

Here's a situation where I'd def prefer async/await. It's not bad, but it can get there easily.

packages/status-screen/src/StatusScreen.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const StatusScreen: FC<{ theme: string }> = ({ theme }) => {
+    const [statusItems, setStatusItem] = useState<SplunkCollectionItem[]>([]);
+
+    useEffect(() => {
+        retrieveTenantId().then((id) => {
+            findCollection().then((items: SplunkCollectionItem[]) => {
+                setStatusItem(
+                    items
+                        .filter((p) => {
```
markkasaboski: p?

packages/status-screen/src/StatusScreen.tsx
```diff
@@ -0,0 +1,87 @@
...
+import './global.css';
+import './StatusScreen.css';
+import { findCollection, retrieveTenantId } from './utils/setupConfiguration';
+import { CollectionKeys, SplunkCollectionItem } from './models/splunk';
+
+const StatusScreen: FC<{ theme: string }> = ({ theme }) => {
+    const [statusItems, setStatusItem] = useState<SplunkCollectionItem[]>([]);
+
+    useEffect(() => {
+        retrieveTenantId().then((id) => {
```
markkasaboski: should we be adding the typing `(id: number) => {}` in this and the following callbacks?

packages/status-screen/src/models/splunk.ts
```diff
@@ -0,0 +1,90 @@
...
+    API_KEY = 'api_key',
+    TENANT_ID = 'tenant_id',
+    INGEST_METADATA_ONLY = 'ingest_metadata_only',
+}
+
+export enum CollectionKeys {
+    START_DATE = 'start_date',
+    LAST_FETCHED = 'timestamp_last_fetch',
+    NEXT_TOKEN = 'next_',
+}
```
markkasaboski: We have to duplicate all of this from package to package? I wonder if we need to create a common package for these interfaces and other things like utils and whatnot.

packages/status-screen/src/utils/configurationFileHelper.ts
```diff
@@ -0,0 +1,166 @@
...
+    );
+    configurationStanzaAccessor = await promisify(configurationStanzaAccessor.fetch)();
+
+    let currentIndex = 'main';
+    if ('index' in configurationStanzaAccessor._properties) {
+        currentIndex = configurationStanzaAccessor._properties.index;
+    }
+
+    return currentIndex;
+}
```
markkasaboski, line 166: common package

markkasaboski (commented):
I know it's a WIP still, but thought I'd get ahead on the review.

I think the main thing is a common package if possible. Reduce the duping. What do you think?

packages/flare/package.json
TyMarc: good point about the flare namespace, i will change it. It needs to be under the dependencies since we would also need it for a production build

packages/status-screen/src/StatusScreen.tsx
TyMarc: for predicate

packages/status-screen/src/StatusScreen.tsx
TyMarc: good point. I will try to find a lint config to make sure it flags them

packages/status-screen/src/models/splunk.ts
TyMarc: That's what I want to do after the release, I tried quickly yesterday but didn't find an obvious way to do it

packages/status-screen/src/StatusScreen.tsx
TyMarc: Yeahhhh you're right, I will try to find the thing that I was talking about yesterday to wait for all the promises instead of stacking them

packages/status-screen/src/StatusScreen.tsx
TyMarc: Which doesn't make sense now that I realize that it's the function name, not the element name. Mind blown, I've called it like that for years

packages/flare/package.json
markkasaboski: Right. I should have phrased the question as "Should these be under dependencies?" since it's currently under devDependencies.

packages/configuration-screen/src/components/ConfigurationCompletedStep.tsx
```diff
@@ -25,7 +25,7 @@ const ConfigurationCompletedStep: FC<{
             </h5>
             <div className="form-group">
                 <div className="button-group">
-                    <Button onClick={() => onBackClicked()} isSecondary>
+                    <Button onClick={(): void => onBackClicked()} isSecondary>
```
markkasaboski: `onBackClicked` sounds weird. Don't have to change it, but thought I'd mention it. I don't think the past tense is necessary. Why not `editConfigurationCallback()` or `onEditConfigurationClick()`?

packages/configuration-screen/src/components/Tooltip.tsx
```diff
@@ -32,10 +32,18 @@ const Tooltip: FC<{}> = ({ children }) => {
 
     return (
         <div className="tooltip-container">
-            <span className="tooltip-button" hidden={!isOpened} onClick={() => toggleOpened(false)}>
+            <span
+                className="tooltip-button"
+                hidden={!isOpened}
+                onClick={(): void => toggleOpened(false)}
```
markkasaboski: Same past tense thing here and others above.

markkasaboski (approved):
LGTM

packages/configuration-screen/src/components/ConfigurationCompletedStep.tsx
TyMarc: You know what, I've never thought about that since I was working with french-speaking devs but that's a really good point

markkasaboski (approved):
LGTM

PR #44, opened 2024-11-15
Fixes (fingers crossed) publish workflow

markkasaboski:


.github/workflows/publish.yml
```diff
@@ -8,12 +8,19 @@ jobs:
...
+      - name: Install Application Dependencies
+        run: |
+          make build
 
       - name: Package and Publish
         run: |
           make package
-          make publish
\ No newline at end of file
+          make publish SPLUNKBASE_CREDS="${{ secrets.SPLUNKBASE_CREDS }}"
```
aviau: usually the env var comes before the command?

`KEY=VALUE some-command`

markkasaboski: It works either way, but if we insist on that convention, I'll change it.

aviau (approved):

PR #48, opened 2024-11-18
Single react components package with multiple entry points instead of multiple packages for each components

TyMarc:


packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -207,6 +229,11 @@ async function fetchCurrentIndexName(): Promise<string> {
     );
 }
 
+async function fetchVersionName(): Promise<string> {
+    const service = createService();
+    return getConfigurationStanzaValue(service, 'app', 'launcher', 'version', 'unknown');
```
markkasaboski, line 234: Should we set the default return value to unkown in the method signature so that we don't have to specify it in the call?

markkasaboski (approved):

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: On fetchVersionName()? I could, I will add it to the stashed changes I have for the status page

PR #54, opened 2024-11-20
Trigger the cron job right after the first configuration

TyMarc:


packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -114,8 +114,30 @@ async function saveConfiguration(
         `${isIngestingMetadataOnly}`
     );
     await saveIndexForIngestion(service, indexName);
+    const isFirstConfiguration = await fetchIsFirstConfiguration();
+    if (isFirstConfiguration) {
+        await updateEventIngestionCronJobInterval(service, 1);
+    }
```
markkasaboski: I wonder if this is necessary right off the top.

What about defaulting the interval to 1 > complete setup > reload app (a job runs instantly) > check isFirstConfiguration > update interval config to 60 > reload app

Does this make sense or am I missing something? I might have derped

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -114,8 +114,30 @@ async function saveConfiguration(
...
+    intervalInSeconds: number
+): Promise<void> {
+    await updateConfigurationFile(
+        service,
+        'inputs',
+        'script://$SPLUNK_HOME/etc/apps/flare/bin/cron_job_ingest_events.py',
+        {
+            interval: `${intervalInSeconds}`,
+        }
+    );
```
markkasaboski: I wonder if we should put a retry mechanism in here should the request fail for whatever reason.

markkasaboski (approved):
A couple of questions. Looks good otherwise.

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: That'd make sense yes

TyMarc: Instead, by default it's scheduled to run once at midnight on january 1st every year until it's configured. At that point it will run once right away and then every minute after the first run is over

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: it might be worth it even if the requests are running locally, wouldn't do it in this PR though.

PR #56, opened 2024-11-25
Make sure we update the last ingested tenant id when switching tenants

TyMarc:


packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -151,8 +151,8 @@ def create_collection(app: client.Application) -> None:
         )
 
 
-def save_start_date(app: client.Application, tenant_id: int) -> None:
-    current_tenant_id = get_current_tenant_id(app=app)
+def save_last_ingested_tenant_id(app: client.Application, tenant_id: int) -> None:
+    last_ingested_tenant_id = get_last_ingested_tenant_id(app=app)
```
markkasaboski: Is there any reason why we'd put this variable here instead of on line 167 where it's used?

Could line 167 be `if get_last_ingested_tenant_id(app=app) != tenant_id`?

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -161,15 +161,21 @@ def save_start_date(app: client.Application, tenant_id: int) -> None:
...
             id=CollectionKeys.START_DATE.value,
             data=json.dumps({"value": date.today().isoformat()}),
         )
 
+        save_collection_value(
+            app=app,
+            key=CollectionKeys.LAST_INGESTED_TENANT_ID.value,
+            value=tenant_id,
+        )
+
```
markkasaboski: Is this meant to be in the if block? Should be outside, no?

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -151,8 +151,8 @@ def create_collection(app: client.Application) -> None:
         )
 
 
-def save_start_date(app: client.Application, tenant_id: int) -> None:
-    current_tenant_id = get_current_tenant_id(app=app)
+def save_last_ingested_tenant_id(app: client.Application, tenant_id: int) -> None:
```
markkasaboski, line 154: This is tricky because this function does two things:
1. saves the started date
2. saves the last_ingested_tenant_id

So how to best name it? 🤔

packages/flare/bin/cron_job_ingest_events.py
TyMarc: I've put it in for the tests to see if it actually changed or not, otherwise it doesn't really matter since the value hasn't changed

packages/flare/bin/cron_job_ingest_events.py
TyMarc: The first save date was useless since on the first run, the last ingested tenant id is None. Any idea how it should be called?

markkasaboski: Aaaah I see, said the blind man when he grabbed his hammer and saw...

markkasaboski: Yeah, dunno about name.. Ugh. naming is the worst. This is fine until someone else raises it. 😂

markkasaboski (approved):
LGTM

PR #57, opened 2024-11-25
Severity and categories filter

TyMarc:
- Moved back the configuration steps business logic into their own component
- Added support for severity and source types in the fetch event requests
- Created components for severity options and a source type list
- Created a constants file
- Block the user from going to the next step if the form isn't valid
- Added Unit Tests for setupConfiguration.ts

https://github.com/user-attachments/assets/9663b095-259e-492e-afed-0a0d46e6e56b


packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -139,7 +147,11 @@ def test_flare_full_data_with_metadata_and_exception(
 
     with pytest.raises(KeyError, match="metadata"):
         for _, _ in flare_api.fetch_feed_events(
```
markkasaboski: Hmm, I know this is unchanged code, but I'm wondering why we bother putting this in a loop. Isn't there only gonna be one call here because we're not getting a next value in the result and passing it in for the subsequent call?

The idea here is to ensure that a call to fetch_feed_events raises a KeyError and with the provided inputs, that's gonna happen on the first call.

packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -100,7 +104,11 @@ def test_flare_full_data_with_metadata(
 
     events: list[dict] = []
     for event, next_token in flare_api.fetch_feed_events(
```
markkasaboski, line 106: Same thing here, really. (I know this is out of order - see comment below first 😂 ) 

Looks like there will only be one call here to me. And then there should be no need for that `events: list[dict] = []`. Am I wrong? I could be wrong. I'm wrong often.

packages/react-components/src/components/ConfigurationGlobalStep.css
```diff
@@ -13,7 +13,7 @@
     width: 100%;
     align-items: flex-start;
     margin-top: 2rem;
-    gap: 1rem;
+    gap: 2rem;
```
markkasaboski, line 16: Aaaaah so much better now 😂

packages/react-components/src/components/ConfigurationInitialStep.tsx
```diff
@@ -1,29 +1,63 @@
...
+                toastManager.show({
+                    id: ToastKeys.ERROR,
+                    isError: true,
+                    content: 'Something went wrong. Please review your form.',
+                });
+            });
+    };
+
+    const isFormValid = (): boolean => {
+        return apiKey.length > 0;
```
markkasaboski, line 52: We don't have to do this, but we could do a couple more checks on the apiKey here:

1. Can ensure that the apiKey starts with `fw_`
2. Can ensure that the total length is 44 chars, including the `fw_`

I'm reasonably certain we don't expect that to change ever.

packages/react-components/src/components/ConfigurationUserPreferencesStep.tsx
```diff
@@ -1,47 +1,191 @@
...
-    isIngestingMetadataOnly,
-    onNavigateBackClick,
-    onSubmitUserPreferencesClick,
-    onTenantIdChange,
-    onIndexNameChange,
-    onIngestingMetadataChange,
-}) => {
+    onUserPreferencesSaved: () => void;
+}> = ({ show, configurationStep, apiKey, onNavigateBackClick, onUserPreferencesSaved }) => {
+    const [tenantId, setTenantId] = useState(-1);
```
markkasaboski: What does the -1 do here? Google is failing me. Rather, I'm probably failing Google.

packages/react-components/src/components/ConfigurationUserPreferencesStep.tsx
```diff
@@ -1,47 +1,191 @@
...
+                        availableSeverities,
+                        availableSourceTypeCategories,
+                        availableIndexNames,
+                        severitiesFilter,
+                        sourceTypesFilter,
+                    ]) => {
+                        setTenantId(id);
+                        setIsIngestingMetadataOnly(ingestMetadataOnly);
+                        setIndexName(index);
+                        if (id === -1 && userTenants.length > 0) {
```
markkasaboski, line 117: I was confused by the `-1`. I see that the `fetchTenantId()` function either returns parsed ID or `-1`. I guess my question is, why not simply `null`?  Null would make sense to me right off the hop, but I had to go and look and figure it out.

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -116,6 +139,8 @@ async function saveConfiguration(
         PasswordKeys.INGEST_METADATA_ONLY,
         `${isIngestingMetadataOnly}`
     );
+    await savePassword(storagePasswords, PasswordKeys.SEVERITIES_FILTER, `${severitiesFilter}`);
```
markkasaboski, line 145: I think we should probably only be storing the API key in storage_passwords. Everything else really should be in KVStore. Thoughts?

packages/flare/bin/flare_external_requests.py
```diff
@@ -20,10 +34,44 @@ def handle_POST(self) -> None:
...
+            raise Exception("API Key is required")
+
+
+class FlareFiltersSeverity(splunk.rest.BaseRestHandler):
+    def handle_POST(self) -> None:
+        logger = Logger(class_name=__file__)
+        payload = self.request["payload"]
+        params = parse.parse_qs(payload)
+
+        if "apiKey" in params:
```
aviau: Instead of indenting the whole body, you probably want to return early?

```python
if "apikey" not in params:
    raise Exception(...)

rest_of_code_here...
```

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -193,16 +218,16 @@ async function fetchPassword(passwordKey: string): Promise<string> {
             return password._properties.clear_password;
         }
     }
-    return '';
+    return undefined;
 }
 
 async function fetchApiKey(): Promise<string> {
-    return fetchPassword(PasswordKeys.API_KEY);
+    return (await fetchPassword(PasswordKeys.API_KEY)) || '';
```
markkasaboski, line 231: Is it necessary to return an empty string? I'm always of the mind that, if you're `get`-ting or `fetch`-ing something, if there is no value, the representation for no value is `null`.

markkasaboski (approved):
Overall, looks good. Mostly questions.

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: That's valid, I'll change it in the next PR; basically the UI is ready once the apiKey value is not undefined anymore, but I want to add a pre-loading step on the configuration screens to avoid this and also show a loading instead of empty fields

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: Yep I will change that

packages/react-components/src/components/ConfigurationUserPreferencesStep.tsx
TyMarc: Years of working with java primitives, the -1 was used by default since we weren't able to set it to null, I'll switch to undefined

packages/react-components/src/components/ConfigurationInitialStep.tsx
TyMarc: Just in case it does change in the future, we already send a call to validate the API key, so I feel like it isn't that 
useful to add UI validation on top

packages/flare/tests/bin/test_flare_wrapper.py
TyMarc: There's going to be two calls here! One for the message 'some_uid' and the other for 'some_other_uid'. They will just return the same full data

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: Thanks to Splunk, we can't do it using their SDK and we use their authentication wrapper to do requests

aviau:
What happens when I select all event types? What will be passed to Flare's API? Is there a tests that check for this?

packages/react-components/src/components/ConfigurationInitialStep.tsx
aviau: > I'm reasonably certain we don't expect that to change ever.

I wouldn't validate the API key. We could definitely change the format in the future.

`fw_` isn't the best prefix now that we don't use `firework` in any public branding anymore

TyMarc:
@aviau If you select every event types or every severities, no filter will be passed in the request; the filter will be deleted from our stored passwords

PR #59, opened 2024-11-26
Moved the built flare folder to the root in an output folder to make the structure more readable

TyMarc:


markkasaboski (approved):
Awesome. Looks good.

PR #64, opened 2024-11-27
Moved the business logic of the configuration steps in the specific steps

TyMarc:
- Added an API key validation to determine if the key is valid instead of fetching the user tenants
- Makefile: No need to check if we have python tests anymore, we do have them
- Removed the splunk.min.js file since it's injected by splunk, removed the compilation of javascript files
- Created constants file for typescript since it was getting all over the place
- Added disabled step for buttons to block the user from submitting before the step is loaded

https://github.com/user-attachments/assets/6d8af421-4873-40d6-874d-fc892556c969


packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -138,9 +142,12 @@ def test_flare_full_data_with_metadata_and_exception(
     flare_api = FlareAPI(api_key="some_key", tenant_id=111)
 
     with pytest.raises(KeyError, match="metadata"):
-        for _, _ in flare_api.fetch_feed_events(
-            next=None, start_date=None, ingest_metadata_only=False
-        ):
-            pass
+        next(
```
markkasaboski, line 145: OOOhhh right! `next()` 👍

packages/react-components/src/ConfigurationScreen.tsx
```diff
@@ -1,156 +1,40 @@
...
+        if (configurationStep === ConfigurationStep.Initial) {
             redirectToHomepage();
-        } else if (currentConfigurationStep === ConfigurationSteps.UserPreferences) {
-            setUserTenants([]);
-        } else if (currentConfigurationStep === ConfigurationSteps.Completed) {
-            reset();
+        } else if (configurationStep === ConfigurationStep.UserPreferences) {
+            setConfigurationStep(ConfigurationStep.Initial);
+        } else if (configurationStep === ConfigurationStep.Completed) {
+            setConfigurationStep(ConfigurationStep.Initial);
```
markkasaboski: I feel like this is a good candidate for a switch. Would clean it up a bit and default should either be the for Initial and `redirectToHomepage()` or thow an unreachable error.

TyMarc: Changed it to a switch and now throwing an error when a new step is added without implementing the back button action!

packages/react-components/src/models/constants.ts
```diff
@@ -0,0 +1,27 @@
+import { SplunkApplicationNamespace } from './splunk';
+
+export const appName: string = 'flare';
+export const storageRealm: string = 'flare_integration_realm';
+export const applicationNameSpace: SplunkApplicationNamespace = {
+    owner: 'nobody',
+    app: appName,
+    sharing: 'app',
+};
+export const flareSavedSearchName = 'Flare Search';
```
markkasaboski: Is there a reason why we don't do all upper case for these constants?

Also, I see typing on some of these but then some not. Not sure if it's even that helpful to include typing for scalar constants. like `const appName = 'flare'` can only ever be a string.

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -57,19 +56,21 @@ function createService(): SplunkService {
...
+function fetchApiKeyValidation(apiKey: string): Promise<boolean> {
     const service = createService();
     const data = { apiKey };
-    service.post('/services/fetch_user_tenants', data, (err, response) => {
-        if (err) {
-            errorCallback(err.data);
-        } else if (response.status === 200) {
-            successCallback(response.data.tenants);
+    return promisify(service.post)('/services/fetch_api_key_validation', data).then(
+        (response: any) => {
```
markkasaboski: I wonder if it's possible to import the actual response type here

markkasaboski (approved):
Couple of nits, but looks good otherwise.

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: Splunk is creating their own object, but I created an interface that has the base variables that we need

PR #66, opened 2024-11-27
Severity filter

TyMarc:
- Fetch available severities via the rest API
    - Added picker in the user preference configuration step
    - Unit tests
    - Logic filtering is:
            - Minimum one severity must be selected
            - If all of the severities are selected, no filter is applied in case we add new severities in the future
            - Otherwise, pass the severities selected as a filter when fetching the events

https://github.com/user-attachments/assets/a3509f03-b58a-41b4-ba47-824b918b8ced

### DIFF
--- a/packages/react-components/src/components/SeverityOptions.tsx
+++ b/packages/react-components/src/components/SeverityOptions.tsx
@@ -1,0 +1,50 @@
+import React, { FC } from 'react';
+
+import { Severity } from '../models/flare';
+import SeverityOption from './SeverityOption';
+import './SeverityOptions.css';
+
+const SeverityOptions: FC<{
+    severities: Severity[];
+    selectedSeverities: Severity[];
+    setSelectedSeverities: (selectedSeverities: Severity[]) => void;
+}> = ({ severities, selectedSeverities, setSelectedSeverities }) => {
+    const isSeverityChecked = (severity: Severity): boolean => {
+        return (
+            selectedSeverities.findIndex(
+                (selectedSeverity) => selectedSeverity.value === severity.value
+            ) >= 0
+        );
+    };
+
+    const handleOnSeverityChange = (severity: Severity, isChecked: boolean): void => {
+        if (isChecked) {
+            setSelectedSeverities([...selectedSeverities, severity]);
+        } else {
+            setSelectedSeverities(
+                selectedSeverities.filter(
+                    (selectedSeverity) => selectedSeverity.value !== severity.value
+                )
+            );
+        }
+    };
+
+    return (
+        <div id="severities-container">
+            {severities.map((severity) => {
+                return (
+                    <SeverityOption
+                        key={severities.indexOf(severity)}
+                        isChecked={isSeverityChecked(severity)}
+                        severity={severity}
+                        onCheckChange={(isChecked): void =>
+                            handleOnSeverityChange(severity, isChecked)
+                        }
+                    />
+                );
+            })}
+        </div>
+    );
+};
+
+export default SeverityOptions;
--- a/packages/react-components/src/tests/setupConfiguration.unit.tsx
+++ b/packages/react-components/src/tests/setupConfiguration.unit.tsx
@@ -1,5 +1,45 @@
-import { getRedirectUrl } from '../utils/setupConfiguration';
+import {
+    convertSeverityFilterToArray,
+    getRedirectUrl,
+    getSeverityFilterValue,
+} from '../utils/setupConfiguration';
+import { getAllSeverities } from './flare.fixture';
 
 test('Flare Redirect URL', () => {
     expect(getRedirectUrl()).toBe('/app/flare');
+});
+
+/** Severity filters */
+test('Select one severity', () => {
+    const allSeverities = getAllSeverities();
+    const selectedSeverities = [allSeverities[0]];
+    const severityFilter = getSeverityFilterValue(selectedSeverities, allSeverities);
+    expect(severityFilter).toBe(allSeverities[0].value);
+});
+
+test('Select no severity', () => {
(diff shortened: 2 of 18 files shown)


packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -77,6 +78,16 @@ function fetchUserTenants(apiKey: string): Promise<Array<Tenant>> {
     );
 }
 
+function fetchFiltersSeverities(apiKey: string): Promise<Array<Severity>> {
+    const service = createService();
+    const data = { apiKey };
+    return promisify(service.post)('/services/fetch_filters_severities', data).then(
```
markkasaboski: Lots of plurals in this. Maybe this makes sense. I dunno.

TyMarc: Yous tells mes, shoulds them?

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -289,19 +311,60 @@ async function fetchVersionName(defaultValue: string): Promise<string> {
     return getConfigurationStanzaValue(service, 'app', 'launcher', 'version', defaultValue);
 }
 
+function convertSeverityFilterToArray(
+    severitiesFilter: string[],
+    availableSeverities: Severity[]
```
markkasaboski: I wonder if it makes sense for this to be called `allSeverities`. I'm OK with `availableSeverities` as well.

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -289,19 +311,57 @@ async function fetchVersionName(defaultValue: string): Promise<string> {
...
+    severitiesFilter.forEach((severityValue) => {
+        if (severityValue === DEFAULT_FILTER_VALUE) {
+            severities.push(...allSeverities);
+        } else {
+            const foundSeverityMatch = allSeverities.find(
+                (severity) => severity.value === severityValue
+            );
+            if (foundSeverityMatch) {
+                severities.push(foundSeverityMatch);
+            }
```
markkasaboski: What about doing this in the the loop? Like 

```
allSeverities.forEach((severity) => {
    if (severity.value === severityValue) {
        severities.push(severity)
    }
});
```

Even better for my eyes because I dislike else blocks would be the following:

```
severitiesFilter.forEach((severityValue) => {
    if (severityValue === DEFAULT_FILTER_VALUE) {
        severities.push(...allSeverities);
        return;
    }

    allSeverities.forEach((severity) => {
        if (severity.value === severityValue) {
            severities.push(severity)
        }
    });
);
 return severities;
```

packages/react-components/src/tests/setupConfiguration.unit.tsx
```diff
@@ -1,5 +1,27 @@
...
+    const selectedSeverities = [];
+    expect(() => getSeverityFilterValue(selectedSeverities, allSeverities)).toThrow(Error);
+});
+
+test('Select all severities', () => {
+    const allSeverities = getAllSeverities();
+    const selectedSeverities = [...allSeverities];
+    const severityFilter = getSeverityFilterValue(selectedSeverities, allSeverities);
+    expect(severityFilter).toBe('');
+});
```
markkasaboski, line 31: I don't know about all of the other functions within setupConfiguration.ts that make network requests, but can we add tests for other functions like `convertSeverityFilterToArray()`?

packages/react-components/src/components/SeverityOptions.tsx
```diff
@@ -0,0 +1,51 @@
...
+            const newSeverities = new Array(...selectedSeverities);
+            newSeverities.push(severity);
+            setSelectedSeverities(newSeverities);
+        } else {
+            const newSeverities = selectedSeverities.filter(
+                (selectedSeverity) => selectedSeverity.value !== severity.value
+            );
+            setSelectedSeverities(newSeverities);
+        }
+    };
```
markkasaboski: We can dedupe some code in here. 

```suggestion
    const handleOnSeverityChange = (severity: Severity, isChecked: boolean): void => {
        if (isChecked) {
           setSelectedSeverities([...selectedSeverities, severity]);
           return;
        } 
        
        setSelectedSeverities(selectedSeverities.filter(
            (selectedSeverity) => selectedSeverity.value !== severity.value
        ));
    };
```

markkasaboski (approved):
A few nits, but looks good otherwise.

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: Not sure to get this one, doing a for loop to find the severity is basically the same as doing a find(), right? Was the problem more regarding the first condition for the default filter value? In the new patchset, I removed this altogether since it was old code to support an empty filter

markkasaboski: For me, it's about fewer lines of code and no need for the variable `foundSeverityMatch`.

markkasaboski: In your latest change, again some contextual spacing goes a long way in terms of readability. Blobbing everything together with no line spacing increases cognitive load. That's largely my opinion. Don't have to change it. I just like code to read like paragraph's instead of run-on sentences and no breaks.

TyMarc: I will separate the code more and add some comments for clarity. The reason I prefer using a find() instead of a forEach is because you can't leave early in a forEach unless you throw an Exception. Would be nice if you could push only if the variable wasn't undefined

markkasaboski: Right! I forgot about that. Yeah, that's fair.

PR #67, opened 2024-11-27
Source types filter

TyMarc:
- Fetch available source types via the rest API
- Added picker in the user preference configuration step
- Unit tests
- Logic filtering is:
        - Minimum one source type must be selected
        - If all of the source types are selected, no filter is applied in case we add new source type categories/sub types in the future
        - If all of the subtypes of a category is selected, the name of the category will be passed in the filter instead of each children in case the sub types are changed in the future
        - Otherwise, pass the source types selected as a filter when fetching the events

https://github.com/user-attachments/assets/e8d10c42-a5f8-4917-8d15-7ee357642490


packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -172,6 +174,18 @@ def get_severities_filter(storage_passwords: StoragePasswords) -> list[str]:
...
+
+    if source_types_filter:
+        return source_types_filter.split(",")
+
+    return []
+
+
 def get_next(kvstore: KVStoreCollections, tenant_id: int) -> Optional[str]:
     return get_collection_value(
         kvstore=kvstore, key=f"{CollectionKeys.get_next_token(tenantId=tenant_id)}"
```
markkasaboski: `CollectionKeys.get_next_token(tenantId=tenant_id)` returns a string and we're not doing any further concatenation, so does this need to be within `f""`?

Right, this is unchanged code. As in, you didn't modify it. But I'm seeing it now and being all like 🤔

TyMarc: Nah you're right, it doesn't need to; since it's the last commit of the chain, i'll modify it!

packages/react-components/src/components/ConfigurationUserPreferencesStep.tsx
```diff
@@ -124,11 +149,17 @@ const ConfigurationUserPreferencesStep: FC<{
...
             setSelectedSeverities([]);
+            setSourceTypeCategories([]);
+            setSelectedSourceTypes([]);
         }
     }, [configurationStep, apiKey]);
 
     const isFormValid = (): boolean => {
-        return tenantId !== undefined && selectedSeverities.length > 0;
+        return (
+            tenantId !== undefined &&
```
markkasaboski, line 159: Did we switch this from -1 to undefined? I think I remember seeing it as -1 still.

TyMarc: In the refactor PR, it was changed to undefined instead of -1

TyMarc: Oh no you're right, i'll change it right away

packages/react-components/src/tests/setupConfiguration.unit.tsx
```diff
@@ -43,3 +45,82 @@ test('Convert severity filter (part of them)', () => {
...
+    const allSourceTypeCategories = getAllSourceTypeCategories();
+    const selectedSourceTypes = [allSourceTypeCategories[0].types[0]];
+    const sourceTypeFilter = getSourceTypesFilterValue(
+        selectedSourceTypes,
+        allSourceTypeCategories
+    );
+    expect(sourceTypeFilter).toBe(allSourceTypeCategories[0].types[0].value);
+});
+
+test('Select one source type', () => {
```
markkasaboski: copy/paste title with the above test?

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -343,6 +364,62 @@ function getSeverityFilterValue(selectedSeverities: Severity[], allSeverities: S
...
+            const foundSourceTypeMatch = allSourceTypeCategories
+                .flatMap((category) => category.types)
+                .find((sourceType) => sourceType.value === sourceTypeValue);
+            if (foundSourceTypeMatch) {
+                sourceTypes.push(foundSourceTypeMatch);
+            }
+        }
+    });
+    return sourceTypes;
+}
```
markkasaboski: This is difficult to read. Some contextual spacing would help a bit. Also, less indentation. Can lose the else block.

markkasaboski (approved):
Looks good

PR #68, opened 2024-11-28
Added a loading animation when initializing the data of a configuration step

TyMarc:
- Return null instead of hidden the HTML elements of a step to avoid useless rendering
- Reset the data of the ConfigurationStep.Completed when re-editing
- Shiny spinning Flare logo when loading the data

https://github.com/user-attachments/assets/bfd6181d-61f0-4d72-8d3b-029aaa1d4dde


packages/react-components/src/components/ConfigurationInitialStep.tsx
```diff
@@ -9,29 +9,29 @@ import { fetchApiKeyValidation } from '../utils/setupConfiguration';
...
 }) => {
+    const [isInitializingData, setIsInitializingData] = useState(true);
     const [errorMessage, setErrorMessage] = useState('');
     const [isLoading, setIsLoading] = useState(false);
     const handleApiKeyChange = (e): void => setApiKey(e.target.value);
 
     const handleSubmitApiKey = (): void => {
         setIsLoading(true);
-        fetchApiKeyValidation(apiKey)
+        fetchApiKeyValidation(apiKey!!)
```
markkasaboski: What's going on here?

TyMarc: Since the apiKey is now undefined by default, lint was indicating that the value could have been undefined, but if we're pressing the submit button, it means that the api key value cannot be undefined, so I force it as a string

markkasaboski (approved):
Looks good. Just one question.

packages/react-components/src/components/ConfigurationInitialStep.tsx
markkasaboski: I've never seen this before. That will force `undefined` to be `'undefined'`? I get an error when I try this. `apiKey!!` Seems like a syntax error to me.

TyMarc: There seems to be a config that makes it work, but it's odd since I don't see it applied. I will switch back to a normal if to check if the variable is undefined

PR #69, opened 2024-11-28
Refactors test_ingest_events.py removing Mocks

markkasaboski:


packages/flare/tests/bin/test_ingest_events.py
```diff
@@ -30,58 +34,54 @@
...
-def test_save_collection_value_expect_insert() -> None:
+def test_save_collection_value_expect_insert(kvstore: FakeKVStoreCollections) -> None:
     key = "some_key"
     value = "some_value"
-    kvstore = MagicMock()
     save_collection_value(kvstore=kvstore, key=key, value=value)
-    kvstore[KV_COLLECTION_NAME].data.insert.assert_called_once_with(
-        json.dumps({"_key": key, "value": value})
-    )
+    assert kvstore[KV_COLLECTION_NAME].data.query() == [{"_key": key, "value": value}]
```
TyMarc: Instead of testing the fake KV store, wouldn't it be better to test that `get_collection_value` returns the proper value? This test and the one before could be merged together to test the save + getter instead

packages/flare/tests/bin/test_ingest_events.py
```diff
@@ -30,58 +34,54 @@
...
+    kvstore.create(name=KV_COLLECTION_NAME, fields={})
+    kvstore[KV_COLLECTION_NAME].data.insert(
+        json.dumps(
+            {"_key": key, "value": "old_value"},
+        )
     )
 
+    assert get_collection_value(kvstore=kvstore, key=key) == "old_value"
+    save_collection_value(kvstore=kvstore, key=key, value=value)
+    assert kvstore[KV_COLLECTION_NAME].data.query() == [{"_key": key, "value": value}]
```
TyMarc, line 56: Same thing here, you could use `get_collection_value` to assert the value

packages/flare/tests/bin/conftest.py
TyMarc, line 1: That's great, we will be able to actually test the collections

packages/flare/tests/bin/test_ingest_events.py
markkasaboski: These tests: `test_get_collection_value_expect_none` & `test_get_collection_value_expect_result` do exectly that. There's no need to use it in every other test. It wouldn't hurt to use it in every other test, but that's not the focus of these other tests. In this test in particular, `save_collection_value()` is the focus and I'd personally prefer to go directly to the source for the assertion rather than using the wrapper around it. ¯\_(ツ)_/¯, but I'm easy. I can be swayed if someone makes a good argument.

packages/flare/tests/bin/test_ingest_events.py
markkasaboski: Ha! But it's funny that I see I used it to assert the old value in this one. It was the end of the day. I don't know what I was thinking at that point. 

I mean, it's certainly cleaner. Looks better. I dunno... I still prefer to go directly to the source for the assertion. I just need someone to force me one way or the other 😂

packages/flare/tests/bin/test_ingest_events.py
TyMarc: Yeah that's fair, the method and the direct query does the same thing so it's actually just a nit

TyMarc (approved):
LGTM, that's great!

PR #73, opened 2024-12-06
Change the option from 'Ingest metadata only' to 'Ingest full event data' and make it off by default

TyMarc:
https://github.com/user-attachments/assets/d53a2256-cc85-4ca5-97b8-584cab301ee2


aviau:
so fastttt

aviau:
I imagine we need a frontend change too? Can you post a screenshot?

TyMarc:
Added video in description @aviau

aviau:
Can we reverse the checkbox?

Name it "Full Event Ingestion", and turn it off by default.

TyMarc:
@aviau I updated the video

aviau (approved):

PR #79, opened 2025-01-15
Rewrote the interfaces of the Splunk object to follow the hierarchy of the Javascript SplunkSDK

TyMarc:
Based on @aviau 's  comment, a wrapper would be better to do better error handling in the future.

Will be used to simplify the creation of a wrapper to generate Promises on top of the JS SDK

### DIFF
--- a/packages/react-components/src/models/splunk.ts
+++ b/packages/react-components/src/models/splunk.ts
@@ -1,103 +1,65 @@
-export interface SplunkApplicationNamespace {
-    app?: string;
-    owner?: string;
-    sharing?: string;
-}
-
-export interface SplunkPassword {
-    name: string;
-    _properties: {
-        clear_password: string;
-    };
-}
-
-export interface SplunkCollectionItem {
+export interface KVCollectionItem {
     key: string;
     value: string;
     user: string;
 }
 
-export interface SplunkIndex {
-    name: string;
+export interface ApplicationNamespace {
+    app?: string;
+    owner?: string;
+    sharing?: string;
 }
 
-export interface SplunkSavedSearch {
-    name: string;
-    qualifiedPath: string;
-    update: (properties: Record<string, string>) => SplunkRequestResponse;
-}
-
-export interface SplunkAppAccessor {
-    reload: () => void;
-}
-
-export interface Stanza {
-    name: string;
-}
-
-export interface SplunkRequestResponse {
+export interface HTTPResponse {
     status: number;
     data: any;
 }
 
-export interface SplunkAppsAccessor {
-    fetch: () => SplunkAppsAccessor;
-    item: (applicationName: string) => SplunkAppAccessor;
+export interface Endpoint {
+    fetch: () => this;
+    del: (relativePath: string) => HTTPResponse;
+    qualifiedPath: string;
 }
 
-export interface ConfigurationStanzaAccessor {
-    fetch: () => ConfigurationStanzaAccessor;
-    update: (properties: Record<string, string>) => SplunkRequestResponse;
-    list: () => Array<{ name: string }>;
+export interface Resource extends Endpoint {
     properties: () => Record<string, string>;
-    _properties: Record<string, string>;
 }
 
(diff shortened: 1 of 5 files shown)


packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -58,44 +58,60 @@ function createService(): SplunkService {
     return service;
 }
 
+function handleHTTPScriptError(error: any): Promise<boolean> {
```
aviau: Would it make sense to extract this in a `SplunkHTTPClient` somewhere?

Otherwise we have to copy paste this handler's pointer everywhere?

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -58,44 +58,60 @@ function createService(): SplunkService {
...
+                    'The Splunk instance needs to be restarted to recognize the scripts of this application.'
+                );
+            } else if (error.data.messages !== undefined && error.data.messages.length > 0) {
+                throw Error(error.data.messages[0].text);
+            }
+            throw Error(error.data);
+        });
+    }
+    throw Error(error.data);
+}
```
markkasaboski: I'm just gonna make a note of this here.

Splunk Cloud cannot be restarted, so if this error is encountered there, it will be misleading and cause confusion. So support will have to be made aware of this.

TyMarc: Updated to check if the current Splunk server is a cloud-based

packages/react-components/src/utils/setupConfiguration.ts
TyMarc: We are actually using the Splunk SDK's HTTP Client, that's why I'm adding the error handling in the Promise directly. Also, the script error handling only apply to the four endpoints we defined in the flare_external_requests script file, not the other HTTP request that are done on the Splunk SDK objects

aviau: So then my suggestion is to wrap the Splunk SDK's client?

TyMarc: Not sure how I would do that, I'm not working directly with their HTTP client, they already have a wrapper above it to expose their interface, they just append to also have the methods `get` and `post` available to do requests with your authentication. The Splunk SDK is injected in the page, it's not a dependency that we have in the project.

aviau: Similar to this?:

```js
const splunkClient = () => {
    const post = (url, data) => Promise {
        return promisify(service.post)(url, data).catch(handleHTTPServiceError)
    }
    return {
        post,
    };
}();
```

TyMarc: Got it. It's not a small refactor, I might finish this PR during the cooldown or if I have downtime during the cycle.

TyMarc:
This PR is now a refactor of the Splunk interface typing on top of the JS SplunkSDK. I will create two other PRs on top of this one: one for the wrapper that will promisify the functions and another one that will actually do the error handling

packages/react-components/src/models/splunk.ts
```diff
@@ -1,103 +1,66 @@
...
+    del: (relativePath: string) => HTTPResponse;
+    qualifiedPath: string;
 }
 
-export interface ConfigurationStanzaAccessor {
-    fetch: () => ConfigurationStanzaAccessor;
-    update: (properties: Record<string, string>) => SplunkRequestResponse;
-    list: () => Array<{ name: string }>;
+export interface Resource extends Endpoint {
+    fetch: () => this;
```
markkasaboski: Does this need to be redeclared here if it's declared in Endpoint that this interface extends from?

markkasaboski (approved):
Just the one question. Otherwise, LGTM

packages/react-components/src/models/splunk.ts
TyMarc: You're right!

PR #80, opened 2025-01-16
Force Python 3.9 since that's what we need for our scripts to run

TyMarc:
If a customer needs to force a custom python version for their application to run, this will make sure our scripts will still run using Python 3.9

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ package: packages/flare/bin/vendor
 
 .PHONY: publish
 publish: output/flare.tar.gz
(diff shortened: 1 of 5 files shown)


aviau (approved):

TyMarc:
Even the new Splunk AppInspect is returning a validation error if you force the python version to python 3.9, even if it's part of the available options: https://docs.splunk.com/Documentation/Splunk/latest/Admin/Restmapconf

>python.version={default|python|python2|python3|python3.7|python3.9|latest}

aviau:
ah RIP

PR #81, opened 2025-02-04
Adds check for python versions below 3.9 and bails out if true

markkasaboski:


TyMarc:
Should we also do it for the REST endpoints?

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -353,6 +353,9 @@ def get_splunk_service(logger: Logger) -> Service:
 
 
 if __name__ == "__main__":
+    if sys.version_info < (3, 9):
```
aviau: I think the right place to do this would be at the top of the file, before any import (except sys)

The reason for this is that the imports might fail on unsupported versions of python.

markkasaboski: Gotcha. Yeah, makes sense.

markkasaboski:
> Should we also do it for the REST endpoints?

Good question. I don't think it's necessary. @aviau ?

aviau (approved):

aviau:
It depends on the load order.

PR #83, opened 2025-02-07
Fixes splunk client authentication issue

markkasaboski:
https://gitlab.com/flaresystems/ticket-tracker/-/issues/1683

If a user encounters a Splunk client authentication issue, they will now be able to find this error message in the Splunk internal log. Though, the likelihood of anyone seeing this error is slim. The account of the user who first configured the app will have to have been deleted.

Screenshot 2025-02-07 at 4 19 18 PM

### DIFF
--- a/packages/react-components/src/utils/setupConfiguration.ts
+++ b/packages/react-components/src/utils/setupConfiguration.ts
@@ -152,6 +152,7 @@ async function saveConfiguration(
         SEVERITY_SAVED_SEARCH_NAME,
         `source=${APP_NAME} index=${indexName} earliest=-24h latest=now | spath path=header.risk.score output=risk_score_str | eval risk_score = coalesce(tonumber(risk_score_str), 0)  | eval risk_label = case(risk_score == 1, "Info", risk_score == 2, "Low", risk_score == 3, "Medium", risk_score == 4, "High", risk_score == 5, "Critical")  | stats count by risk_label, risk_score | sort risk_score | fields - risk_score`
     );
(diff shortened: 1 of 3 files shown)


TyMarc (approved):

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -174,6 +175,19 @@ async function updateEventIngestionCronJobInterval(
     );
 }
 
+async function updatePassAuthUsername(service: Service): Promise<void> {
+    const username = await fetchCurrentUsername();
+    console.log(username);
```
xvaier: Is this left here for long term logging purposes? 👀

packages/react-components/src/utils/setupConfiguration.ts
```diff
@@ -422,6 +436,13 @@ function getSourceTypesFilterValue(
     return sourceTypesFilter;
 }
 
+function fetchCurrentUsername(): Promise<string> {
+    const service = createService();
+    return promisify(service.currentUser)().then((user) => {
```
xvaier, line 440: This is an interesting pattern, why not just declare the function as async?

xvaier (approved):
This looks like a good fix, maybe remove the console log of the username, but otherwise lgtm

packages/react-components/src/utils/setupConfiguration.ts
markkasaboski: Ha! This one line is the ticket to the entire thing. Without it, it all comes crashing down 😂

PR #87, opened 2025-02-10
Fixes Makefile publish target  credential issue

markkasaboski:
**Problem**

We changed the Splunk app ownership from the Splunk account `integrations@flare.systems` to the Splunk account `integrations+splunk@flare.io`. This means that we have to update the SPLUNK_CREDS Github secret with the new credentials. The password in these new credentials contained the characters `$^` . When these credentials were passed into the Makefile, Make would substitute the `$^` with the Makefile target's first dependency. So the curl would end up like the following:

Screenshot 2025-02-10 at 2 34 27 PM

Obviously, the publish failed.

**Solution**
One solution is to change the password removing those special characters, but this is fragile because it could happen again in the future. 

The solution was as follows:

Instead of doing: `curl -u "$(SPLUNKBASE_CREDS)" ...` 

We do: `curl -u $$SPLUNKBASE_CREDS ...` 

We also had to alter the publish.yml Github workflow to use single quotes instead of double as the double quotes would result in everything from the $ onward from being omitted. Bash was consuming it.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,4 +23,4 @@ jobs:
       - name: Package and Publish
         run: |
           make package
-          SPLUNKBASE_CREDS="${{ secrets.SPLUNKBASE_CREDS }}" make publish
+          SPLUNKBASE_CREDS='${{ secrets.SPLUNKBASE_CREDS }}' make publish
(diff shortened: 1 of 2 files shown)


TyMarc (approved):
LGTM

PR #90, opened 2025-02-15
Sets asc order on event request and start_date is now - 30 days

markkasaboski:
30 days before today:

Screenshot 2025-02-15 at 5 08 12 PM

### DIFF
--- a/packages/flare/bin/cron_job_ingest_events.py
+++ b/packages/flare/bin/cron_job_ingest_events.py
@@ -8,8 +8,9 @@ if sys.version_info < (3, 9):
 import json
 import os
 
-from datetime import date
 from datetime import datetime
+from datetime import timedelta
+from datetime import timezone
 from typing import Any
 from typing import Iterator
 from typing import Optional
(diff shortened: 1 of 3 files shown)


packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -255,7 +256,7 @@ def save_last_ingested_tenant_id(kvstore: KVStoreCollections, tenant_id: int) ->
         save_collection_value(
             kvstore=kvstore,
             key=CollectionKeys.START_DATE.value,
-            value=date.today().isoformat(),
+            value=(date.today() - timedelta(days=30)).isoformat(),
```
aviau: I would prefer if we always used dates with timezones:
```python
(datetime.now(timezone.utc) - timedelta(days=30)).isoformat()
```

markkasaboski: Are you OK with `date.now()`? We have no use for the time component.

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -10,6 +10,8 @@
 
 from datetime import date
```
aviau: Why do we still use `date` in this file?

Everything should be a `datetime`.

I shoud have caught this earlier, sorry.

aviau (approved):

PR #94, opened 2025-02-19
handle errors while fetching full events

markkasaboski:
It seems that we sometimes get no activity data back from the `firework/v2/activities/<uid>` endpoint. I added some logging in to spot this:
Screenshot 2025-02-20 at 10 41 22 AM
Screenshot 2025-02-20 at 10 41 31 AM
Screenshot 2025-02-20 at 10 41 41 AM

This seems to happen fairly often. We get a next token, but nothing for the data. Assuming this is to be expected and to prevent the Splunk app from sending {} to the index, I've added a check to prevent this. 

@aviau Does this make sense and are you OK with the fix? Don't read too much into `Failed to fetch event with uid:`. That's just my logging. The request succeeded.

We get a 500:
Screenshot 2025-02-20 at 1 02 57 PM

I'm thinking we need retries here.

### DIFF
--- a/packages/flare/bin/flare.py
+++ b/packages/flare/bin/flare.py
@@ -47,7 +47,11 @@ def get_flare_api_client(
 
 class FlareAPI(AuthBase):
     def __init__(
-        self, *, api_key: str, tenant_id: Optional[int] = None, logger: Logger
+        self,
+        *,
+        api_key: str,
+        tenant_id: Optional[int] = None,
+        logger: Logger,
     ) -> None:
         self.flare_client = get_flare_api_client(
             api_key=api_key,
@@ -117,8 +121,8 @@ class FlareAPI(AuthBase):
             time.sleep(1)
 
     def _fetch_full_event_from_uid(self, *, uid: str) -> dict:
-        NUMBER_OF_RETRIES = 3
-        for current_try in range(NUMBER_OF_RETRIES):
+        number_of_retries = 3
+        for current_try in range(number_of_retries):
             try:
                 event_response = self.flare_client.get(
                     url=f"/firework/v2/activities/{uid}"
@@ -127,12 +131,12 @@ class FlareAPI(AuthBase):
             except Exception as e:
                 time.sleep(1)
                 self.logger.info(
-                    f"Failed to fetch event {current_try + 1}/{NUMBER_OF_RETRIES} retries: {e}"
+                    f"Failed to fetch event {current_try + 1}/{number_of_retries} retries: {e}"
                 )
                 continue
             return event_response.json()["activity"]
         raise Exception(
-            f"failed to fetch full event data for {uid} after {NUMBER_OF_RETRIES} tries"
+            f"failed to fetch full event data for {uid} after {number_of_retries} tries"
         )
 
     def fetch_api_key_validation(self) -> requests.Response:
--- a/packages/flare/bin/flare.py
+++ b/packages/flare/bin/flare.py
@@ -76,7 +76,7 @@ class FlareAPI(AuthBase):
                     event = self._fetch_full_event_from_uid(
                         uid=event["metadata"]["uid"]
                     )
-                    time.sleep(1)
+                    time.sleep(1)  # Don't hit rate limit
                 yield (event, next_token)
 
     def _fetch_event_feed_metadata(
@@ -115,10 +115,20 @@ class FlareAPI(AuthBase):
             time.sleep(1)
 
     def _fetch_full_event_from_uid(self, *, uid: str) -> dict:
-        event_response = self.flare_client.get(url=f"/firework/v2/activities/{uid}")
-        event = event_response.json()["activity"]
-        self.logger.debug(event)
-        return event
+        for current_try in range(3):
+            try:
+                event_response = self.flare_client.get(
+                    url=f"/firework/v2/activities/{uid}"
+                )
+                event_response.raise_for_status()
+            except Exception as e:
+                time.sleep(1)
+                self.logger.debug(f"Failed to fetch event: {e}")
+                continue
+            return event_response.json()["activity"]
+        raise Exception(
+            f"failed to fetch full event data for {uid} after {current_try + 1} tries"
+        )
 
     def fetch_api_key_validation(self) -> requests.Response:
         return self.flare_client.get(
--- a/packages/flare/bin/flare.py
+++ b/packages/flare/bin/flare.py
@@ -46,12 +46,14 @@ def get_flare_api_client(
 
 
 class FlareAPI(AuthBase):
-    def __init__(self, *, api_key: str, tenant_id: Optional[int] = None) -> None:
+    def __init__(
+        self, *, api_key: str, tenant_id: Optional[int] = None, logger: Logger
+    ) -> None:
         self.flare_client = get_flare_api_client(
             api_key=api_key,
             tenant_id=tenant_id,
         )
-        self.logger = Logger(class_name=__file__)
+        self.logger = logger
 
     def fetch_feed_events(
         self,
@@ -115,7 +117,8 @@ class FlareAPI(AuthBase):
             time.sleep(1)
 
     def _fetch_full_event_from_uid(self, *, uid: str) -> dict:
-        for current_try in range(3):
+        NUMBER_OF_RETRIES = 3
+        for current_try in range(NUMBER_OF_RETRIES):
             try:
                 event_response = self.flare_client.get(
                     url=f"/firework/v2/activities/{uid}"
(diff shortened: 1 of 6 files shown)


packages/flare/bin/flare.py
```diff
@@ -116,8 +116,11 @@ def _fetch_event_feed_metadata(
 
     def _fetch_full_event_from_uid(self, *, uid: str) -> dict:
         event_response = self.flare_client.get(url=f"/firework/v2/activities/{uid}")
-        event = event_response.json()["activity"]
```
xvaier, line 119: This changes what the method returns, are the places where it is used expecting the full event response rather than just the activity?

markkasaboski: I forgot the `["activity"]`. Just added it back. 

It's kind of funny. The very first thing that came into my head when my eyes opened this morning was "I think I forgot to reference 'activity'". Sure enough, I immediately got up and opened my computer and there was your comment 😂

xvaier (dismissed):

packages/flare/bin/flare.py
```diff
@@ -73,10 +73,25 @@ def fetch_feed_events(
             next_token = event_feed["next"]
             for event in event_feed["items"]:
                 if ingest_full_event_data:
-                    event = self._fetch_full_event_from_uid(
-                        uid=event["metadata"]["uid"]
-                    )
-                    time.sleep(1)
+                    uid = event["metadata"]["uid"]
+                    retries = 3
+                    for attempt in range(retries):
```
aviau: Can we hide that logic in the `_fetch_full_event_from_uid` function?

Its not clear to the reader in which scenarios the sub-function will raise exceptions. So it would be easier to read if the exception was both raised and caught inside `_fetch_full_event_from_uid`.

packages/flare/bin/flare.py
```diff
@@ -73,10 +73,25 @@ def fetch_feed_events(
             next_token = event_feed["next"]
             for event in event_feed["items"]:
                 if ingest_full_event_data:
-                    event = self._fetch_full_event_from_uid(
-                        uid=event["metadata"]["uid"]
-                    )
-                    time.sleep(1)
+                    uid = event["metadata"]["uid"]
+                    retries = 3
```
aviau: I think we can declare that variable in the for loop?

```python
for attempt in range(3):
    ....
```

markkasaboski: It is for this check unless we want to hard code 3 in several spots:

```
if attempt < retries - 1:
    time.sleep(1)  # Don't hit rate limit
else:
    self.logger.error(
        f"Failed to fetch full event data for UID {uid} after {retries} attempts."
    )
    raise e
```

packages/flare/bin/flare.py
markkasaboski: Can do. Though, I just want to point out that this is the opposite of what you mentioned in slack thread. Just want to avoid confusion.

packages/flare/bin/flare.py
aviau: Yeah I don't really like the fact that we have to implement retry logic.

I guess the thing I want to avoid is `if attempt < retries - 1:` is hard to read for me. Comparing loop indices is tricky, you always end up doing the wrong thing lol.

It will feel a bit simpler when you hide that complexity in the fech function, something like this?

```python
def fetch_event(uid: str) -> dict:
    
    for current_try in range(3):
        try:
            event_resp = client.get(...)
            event_resp.raise_for_status()
            return event_resp.json()['activity']
        except Exception:
          sleep()
          continue
    raise Exception("failed to fetch full event data for {uid} after {current_try} tries")
```

Downside with this is that I don't include the original exception.

packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -157,3 +158,48 @@ def test_flare_full_data_with_metadata_and_exception(
...
+        "next": "some_next_value",
+        "items": [
+            {
+                "metadata": {"uid": "some_other_uid"},
+                "data": "bar",
+            },
+        ],
+    }
+
+    response_mock = Mock()
```
aviau: Is this the right tool for the job? I feel like this test would be much easier to read with requests-mock?

https://pypi.org/project/requests-mock/

packages/flare/bin/flare.py
aviau: In that thread, I wasn't expect you to implement retries at all. I was just expecting to let it die.

aviau:
Can we set a more meaningful pull request title here?

Something like:
- `handle errors while fetching full events` ?

packages/flare/tests/bin/test_flare_wrapper.py
markkasaboski: Yeah, it's kind of evolved from the activity KeyError, though, still related.

markkasaboski:
> Can we set a more meaningful pull request title here?
> 
> Something like:
> 
> * `handle errors while fetching full events` ?

Yeah, it's kind of evolved since the activity KeyError, but still related.

requirements.tools.txt
```diff
@@ -4,3 +4,5 @@ ruff==0.7.0
 splunk-appinspect==3.9.1
 isort==5.13.2
 freezegun==1.5.1
+
```
aviau: extra empty line here?

packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -157,3 +159,73 @@ def test_flare_full_data_with_metadata_and_exception(
...
+    type(response_mock).status_code = PropertyMock(return_value=200)
+    response_mock.json.return_value = expected_return_value
+    fetch_event_feed_metadata_mock.return_value = iter([response_mock])
+
+    flare_api = FlareAPI(api_key="some_key", tenant_id=111)
+    flare_api.logger = MagicMock()  # logger refuses to patch ¯\_(ツ)_/¯
```
aviau: Why do we need to replace the logger? Could FlareAPI have a parameter that sets the logger we want? Or how about we just let the logging happen? Is it because you want to assert the exact logged messages? It that really necessary?

packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -157,3 +159,73 @@ def test_flare_full_data_with_metadata_and_exception(
...
+        "next": "some_next_value",
+        "items": [
+            {
+                "metadata": {"uid": "some_uid"},
+                "data": "some_data",
+            },
+        ],
+    }
+
+    response_mock = Mock()
```
aviau: Could we also use requests-mock to mock that response?

packages/flare/tests/bin/test_flare_wrapper.py
markkasaboski: Validating the logging is a good and easy way to assert what's actually happening step by step within that retry code. But will take other suggestions.

I don't know why, but patching the logger doesn't work. So yeah, either this or adding it as a constructor param is fine.

aviau: In this case here, you can easily view everything that is happening by looking at http requests?

However if you really want to assert all logs, I think I would prefer to pass a fake logger instead of mocking? We should try to keep mocking to a minimum for when writing testable interfaces is really not possible.

packages/flare/bin/flare.py
```diff
@@ -46,12 +46,14 @@ def get_flare_api_client(
 
 
 class FlareAPI(AuthBase):
-    def __init__(self, *, api_key: str, tenant_id: Optional[int] = None) -> None:
+    def __init__(
+        self, *, api_key: str, tenant_id: Optional[int] = None, logger: Logger
```
aviau: Don't forget  your  trailing commas ;p

packages/flare/bin/flare.py
```diff
@@ -115,10 +117,23 @@ def _fetch_event_feed_metadata(
             time.sleep(1)
 
     def _fetch_full_event_from_uid(self, *, uid: str) -> dict:
-        event_response = self.flare_client.get(url=f"/firework/v2/activities/{uid}")
-        event = event_response.json()["activity"]
-        self.logger.debug(event)
-        return event
+        NUMBER_OF_RETRIES = 3
```
aviau: dont use allcaps if it isn't a global?

packages/flare/tests/bin/test_flare_wrapper.py
```diff
@@ -1,159 +1,256 @@
...
-    with pytest.raises(KeyError, match="metadata"):
-        next(
-            flare_api.fetch_feed_events(
-                next=None,
-                start_date=None,
-                ingest_full_event_data=True,
-                severities=[],
-                source_types=[],
+    with requests_mock.Mocker() as mocker:
+        mocker.register_uri(
```
aviau, line 167: You can extract this in a fixture?

aviau:
Thanks for the refactor. Maybe open a follow-up PR where you extract requests-mock reuses in a fixture?

aviau (approved):

packages/flare/tests/bin/test_flare_wrapper.py
xvaier: Yeah I thought of that, especially for the token route. The other ones change a little in each test.

PR #95, opened 2025-02-21
Fixes lint warnings for the .ruff.toml file settings

markkasaboski:
Screenshot 2025-02-20 at 3 14 53 PM


aviau:
Instead of putting `lint` in every property,  I think you can just put these under a `[lint]` section?

So maybe:
```toml
[lint]

extend-select = [ ]
```

aviau (approved):

PR #96, opened 2025-02-21
Fixes 'Line breaking regex has no capturing groups' error

markkasaboski:
We get a lot of these errors in the internal index:

Screenshot 2025-02-21 at 3 12 22 PM
 
Applying the changes suggested [here](https://community.splunk.com/t5/Getting-Data-In/Why-does-linebreaking-regex-have-no-capturing-groups/m-p/643941/highlight/true#M109661) fixes the issue and events are still ingested and parsed as expected.

Screenshot 2025-02-21 at 4 01 51 PM

@TyMarc Would especially like to get your eyes on this since you created the custom flare_json source. Thanks.


TyMarc (approved):
LGTM!

PR #97, opened 2025-03-04
Move local splunk to a dockerized setup

xvaier:
In order to facilitate local development and reduce globally installed dependencies.

- Moved the local splunk install to a dockerized setup and automated the cache configuration.
- Removed the symlink workflow.


splunk/default.yml
```diff
@@ -0,0 +1,9 @@
+splunk:
```
TyMarc, line 1: That's neat

TyMarc (approved):
LGTM

splunk/default.yml
xvaier: 😎

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    image: splunk/splunk:latest
+    container_name: splunk
+    platform: linux/amd64
+    restart: unless-stopped
+    ports:
+      - "8000:8000"
+    environment:
+      - SPLUNK_START_ARGS=--accept-license
+      - SPLUNK_PASSWORD=a_password
+      - IS_LOCAL_BUILD=true
```
markkasaboski: My only feedback here is a nit. Feel free to ignore.

Typically these type of env vars would be `ENV=dev`, `ENV=prod`, `ENV=test` so on. In this case, I know this only controls the logging, but I'd still prefer to see it as `ENV` and in the logger :

```
if (os.environ.get('ENV') === 'dev') {
    self._logger.setLevel(logging.DEBUG); < instead of INFO
}
```

But whatevs. We can leave it like this as well. I know what to do when I'm in there.

markkasaboski (approved):
👍

docker-compose.yml
xvaier: I agree, I didn't like the `IS_LOCAL_BUILD` variable as I was creating it 😓 

Changed it to `ENV`.

PR #98, opened 2025-03-07
Remove dependency on splunk key value store.

xvaier:
The key value store is an embedded mongo instance and based on summary research (ie  [this forum post](https://community.splunk.com/t5/Monitoring-Splunk/Where-can-I-disable-kvstore/m-p/656006)), not all splunk instances have it enabled.

It also does not run under emulation in docker on Mac, since versions of Splunk after 9.4 use MongoDB 7 that requires AVX instructions.

I decided knowing this information and given that using mongo for a couple of key value lookups is way overkill, to store our app state in an conf file within our app.

I have manually tested that it works and all tests pass.


.gitignore
```diff
@@ -20,4 +20,5 @@ licenses.json
 functional-temp
 splunktional-temp
 
-output/
\ No newline at end of file
+output/
+logs/
```
aviau: It looks like your editor forgets newlines at end of files?

(I see this in a couple of the other files too)

aviau:
Does this support migrating for existing users? (its fine if the answer is no)

.gitignore
xvaier: Yeah I saw that too, I don't know why this setting was disabled. Just enabled it.

packages/flare/bin/data_store.py
```diff
@@ -0,0 +1,110 @@
...
+config_path = os.path.join(
+    splunk_home, "etc", "apps", f"{APP_NAME}", "local", "data_store.conf"
+)
+
+
+class ConfigDataStore:
+    _instance = None
+    _store: configparser.ConfigParser
+
+    # Make ConfigDataStore a singleton
```
aviau: Why is it a singleton?

packages/flare/src/main/resources/splunk/default/restmap.conf
```diff
@@ -17,3 +17,9 @@ python.version = python3
 match=/fetch_source_type_filters
 handler=flare_external_requests.FlareSourceTypeFilters
 python.version = python3
+
+[script:flare_external_requests_ingestion_status]
+match=/fetch_ingestion_status
+handler=flare_external_requests.FlareIngestionStatus
+python.version = python3
+
```
aviau: extra empty line here

xvaier:
> Does this support migrating for existing users? (its fine if the answer is no)

Not automatically, unless we keep some of the kv store logic during the migration and remove it in a second update. I didn't think that was worth the hassle.

packages/flare/src/main/resources/splunk/default/web.conf
```diff
@@ -13,3 +13,7 @@ methods=POST
 [expose:flare_external_requests_source_type_filters]
 pattern=fetch_source_type_filters
 methods=POST
+
+[expose:flare_external_requests_ingestion_status]
+pattern=fetch_ingestion_status
+methods=GET
```
aviau: missing newline here?

packages/flare/bin/data_store.py
xvaier: To avoid having two instances at the same time having inconsistent values and avoiding reading from the file before each get, but since the python scripts are probably run isolated from each other in splunk this might serve no purpose and I might have to read from the file instead of in memory.

packages/flare/bin/flare_external_requests.py
```diff
@@ -81,3 +82,22 @@ def handle_POST(self) -> None:
         logger.debug(f"FlareSourceTypeFilters: {response_json}")
         self.response.setHeader("Content-Type", "application/json")
         self.response.write(json.dumps(response_json))
+
+
+class FlareIngestionStatus(splunk.rest.BaseRestHandler):
```
aviau, line 87: I assume thats not necessarily related to this PR?

So this PR does two things?
- Replace the storage
- Add a new feature of showing status

Should we split this in two PRs? (this is pretty big already)

xvaier: It is related, the previous status was reading from the KV store directly from the frontend without going through a handler.

packages/flare/bin/data_store.py
aviau: Yeah my understanding of the execution model is that there is no two threads in the same process so the singleton just adds complexity and will never be used.

(overall singletons are almost always a red flag)

packages/flare/bin/flare_external_requests.py
aviau: Ahh, thanks for the context

packages/flare/bin/data_store.py
xvaier: Yeah I typically try to avoid them, it will be simpler to add a _sync method that reloads the conf file into memory before getting a key.

packages/flare/bin/data_store.py
```diff
@@ -0,0 +1,110 @@
...
+            "metadata",
+            DataStoreKeys.TIMESTAMP_LAST_FETCH.value,
+            last_fetch.isoformat(),
+        )
+        self._commit()
+
+    def get_next_by_tenant(self, tenant_id: int) -> Optional[str]:
+        return self._store.get(
+            "next_tokens",
+            DataStoreKeys.get_next_token(tenantId=tenant_id),
```
aviau: Looks like this method existing before your PR, but the `tenantId` arg here should be `tenant_id` in snake_case. Do you want to fix it while we are here?

packages/flare/tests/bin/test_data_store.py
```diff
@@ -0,0 +1,26 @@
...
+from datetime import datetime
+
+
+def test_get_and_set_last_tenant_id(data_store: ConfigDataStore) -> None:
+    data_store.set_last_tenant_id(123)
+    assert data_store.get_last_tenant_id() == 123
+
+
+def test_get_and_set_start_date(data_store: ConfigDataStore) -> None:
+    date = datetime(2024, 3, 6, 12, 0, 0)
```
aviau: Oh no this is an evil datetime without timezone!

I guess its fine here since its a test, but these dates are very dangerous because they will cause crashes when compared with timezoned dates.

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -108,23 +50,33 @@ def main(
     severities_filter = get_severities_filter(storage_passwords=storage_passwords)
     source_types_filter = get_source_types_filter(storage_passwords=storage_passwords)
 
-    save_last_fetched(kvstore=kvstore)
-    save_last_ingested_tenant_id(kvstore=kvstore, tenant_id=tenant_id)
+    data_store.set_last_fetch(datetime.now())
```
aviau: Don't create dates without timezones if possible

packages/flare/bin/data_store.py
xvaier: Done ✅

packages/flare/bin/data_store.py
xvaier: Done ✅

aviau:
Maybe you should create a `datetime_now` util somewhere that return `datetime.datetime.now(timezone.utc)` so you don't forget to create timezoned dates?

packages/flare/bin/data_store.py
aviau: You could, but you don't need to. This script isn't run in parallel. You can just remove the singleton and everything will work fine AFAIK?

(but yeah I would probably even remove the state from the class and just read the file every time you want a value, after all this file will be super small)

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -108,23 +50,33 @@ def main(
...
         api_key=api_key,
         tenant_id=tenant_id,
         ingest_full_event_data=ingest_full_event_data,
         severities=severities_filter,
         source_types=source_types_filter,
         flare_api_cls=flare_api_cls,
+        data_store=data_store,
     ):
-        save_last_fetched(kvstore=kvstore)
+        data_store.set_last_fetch(datetime.now())
```
aviau: timezone!

aviau:
Few small comments but overall LGTM for the direction!

packages/flare/bin/data_store.py
```diff
@@ -0,0 +1,110 @@
...
+
+# Define the config file path
+splunk_home = os.environ.get("SPLUNK_HOME", "/opt/splunk")
+config_path = os.path.join(
+    splunk_home, "etc", "apps", f"{APP_NAME}", "local", "data_store.conf"
+)
+
+
+class ConfigDataStore:
+    _store: configparser.ConfigParser
```
aviau: We don't really do that in python.

If you want to define the type for the field, you can do it in the constructor.

```
def __init__(self) -> None:
    self.field: int = 1
```

packages/flare/bin/constants.py
```diff
@@ -18,11 +17,11 @@ class PasswordKeys(Enum):
...
 
 
-class CollectionKeys(Enum):
+class DataStoreKeys(Enum):
     LAST_INGESTED_TENANT_ID = "last_ingested_tenant_id"
     START_DATE = "start_date"
     TIMESTAMP_LAST_FETCH = "timestamp_last_fetch"
 
     @staticmethod
-    def get_next_token(tenantId: int) -> str:
```
markkasaboski, line 27: Ha! We were obviously mixing up our javascripts with our pythons here.

aviau (approved):

packages/react-components/src/models/splunk.ts
```diff
@@ -1,18 +1,12 @@
...
-}
-
 export interface ApplicationNamespace {
     app?: string;
     owner?: string;
     sharing?: string;
 }
 
-export interface HTTPResponse {
+export interface HTTPResponse<T = any> {
```
markkasaboski, line 7: Just a question here: What's the benefit?

xvaier: Oh, there isn't any in this particular case, I went down a rabbit hole of trying to add more dynamic typing and forgot to remove it.

It would allow you to type the data property of the response by using generics instead of casting or hinting elsewhere in the code.

markkasaboski (approved):

PR #99, opened 2025-03-07
Add multi tenancy support

xvaier:
This PR aims to move from a single tenant install to a multi-tenant install with tenant selection in the configuration screen. It will not be backwards compatible with existing installs, but neither was the branch it was based on.

## Changes 
- Added a tenant selector that supports multiple tenants.
- Added support for multiple tenant cursors and ingestion start dates.
- Modified the cron job to support multiple tenants.
- Added a new password key that supports a list of tenants.

### Configuration interface
Screenshot 2025-03-07 at 1 08 35 PM

### Example data storage
```
[tenant_data]
timestamp_earliest_ingested_123 = 2025-03-07T00:00:00.857510+00:00
next_123 =ABCD
timestamp_earliest_ingested_321 = 2025-02-05T17:38:16.020474+00:00
next_321 =ABASED

[metadata]
timestamp_last_fetch = 2025-03-07T18:06:38.586505+00:00
```

### DIFF
--- a/packages/flare/bin/cron_job_ingest_events.py
+++ b/packages/flare/bin/cron_job_ingest_events.py
@@ -52,9 +52,11 @@ def main(
 
     data_store.set_last_fetch(datetime.now(timezone.utc))
 
-    events_fetched_count = 0
+    total_events_fetched_count = 0
 
     for tenant_id in tenant_ids:
+        events_fetched_count = 0
+
         # The earliest ingested date serves as a low water mark to look
         # for identifiers 30 days prior to the day a tenant was first configured.
         start_date = data_store.get_earliest_ingested_by_tenant(tenant_id)
@@ -83,8 +85,10 @@ def main(
             print(json.dumps(event), flush=True)
 
             events_fetched_count += 1
+        logger.info(f"Fetched {events_fetched_count} events on tenant {tenant_id}")
+        total_events_fetched_count += events_fetched_count
 
-    logger.info(f"Fetched {events_fetched_count} events")
+    logger.info(f"Fetched {events_fetched_count} events across all tenants")
 
 
 def fetch_feed(
--- a/packages/flare/bin/cron_job_ingest_events.py
+++ b/packages/flare/bin/cron_job_ingest_events.py
@@ -43,7 +43,7 @@ def main(
         return
 
     api_key = get_api_key(storage_passwords=storage_passwords)
-    tenant_id = get_tenant_id(storage_passwords=storage_passwords)
+    tenant_ids = get_tenant_ids(storage_passwords=storage_passwords)
     ingest_full_event_data = get_ingest_full_event_data(
         storage_passwords=storage_passwords
     )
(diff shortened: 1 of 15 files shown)


packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -52,38 +52,69 @@ def main(
...
+    api_key: str,
+    tenant_id: int,
+    ingest_full_event_data: bool,
+    severities: list[str],
+    source_types: list[str],
+    flare_api_cls: type[FlareAPI],
+    data_store: ConfigDataStore,
+) -> Iterator[tuple[dict, str]]:
+    try:
+        flare_api: FlareAPI = flare_api_cls(api_key=api_key, tenant_id=tenant_id)
```
aviau: Can we move this out of the `try`?

packages/react-components/src/components/ConfigurationUserPreferencesStep.tsx
```diff
@@ -178,17 +175,20 @@ const ConfigurationUserPreferencesStep: FC<{
...
-                            return (
-                                <option key={tenants.indexOf(tenant)} value={tenant.id}>
-                                    {tenant.name}
-                                </option>
-                            );
-                        })}
-                    </Select>
-                    <small className="note">You can only monitor one tenant at a time.</small>
+                    <Label>Tenants</Label>
+                    <TenantSelection
```
aviau, line 179: This could show tenants that are configured but not in flare

aviau:
Can we show the tenant id in the dropdown?

markkasaboski:
I think there are still some unknowns around indexes and how Splunk manages them with regard to their sizes. My fear at the moment is that Orgs with many tenants can fill the index relatively quickly. 

According to the following and what I've seen elsewhere, Splunk stores data in hot, warm, cold and frozen buckets. The frozen bucket, if no path is set, data that is moved here is deleted. If a path is set, it's retained, but it will likely depend on sysadmins to do something with it when this happens. 

Screenshot 2025-03-07 at 2 43 18 PM


This is an area that we have no testing around. Maybe it's OK that that older data is deleted? We may want to at least  mention it and possibly provide some documentation around it. 

`index=_internal sourcetype=splunkd component=BucketMover freeze` < I found that this search query will return log entries of when data is being moved into the frozen bucket. Source: https://www.youtube.com/watch?v=meQZpscK9g0

xvaier:
> I think there are still some unknowns around indexes and how Splunk manages them with regard to their sizes. My fear at the moment is that Orgs with many tenants can fill the index relatively quickly.
> 
> According to the following and what I've seen elsewhere, Splunk stores data in hot, warm, cold and frozen buckets. The frozen bucket, if no path is set, data that is moved here is deleted. If a path is set, it's retained, but it will likely depend on sysadmins to do something with it when this happens.
> 
> `index=_internal sourcetype=splunkd component=BucketMover freeze` < I found that this search query will return log entries of when data is being moved into the frozen bucket. Source: https://www.youtube.com/watch?v=meQZpscK9g0

I think this is a valid concern, we could also create per tenant indexes if this becomes an issue.

packages/react-components/src/components/ConfigurationUserPreferencesStep.tsx
xvaier: image
Done ✅

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -52,38 +52,69 @@ def main(
...
-        # are ingested after being retrieved from Flare.
-        print(json.dumps(event), flush=True)
+            # stdout is picked up by splunk and this is how events
+            # are ingested after being retrieved from Flare.
+            print(json.dumps(event), flush=True)
 
-        events_fetched_count += 1
+            events_fetched_count += 1
 
     logger.info(f"Fetched {events_fetched_count} events")
```
markkasaboski: I'm thinking it would be good to these counts logged on a per-tenant basis as well. Happy to be overruled though.

xvaier: Yeah I was thinking the same thing as I wrote it, I'll change it!

xvaier: Done!

packages/flare/bin/data_store.py
```diff
@@ -16,14 +16,14 @@
...
         config_store.read(config_path)
 
         # Add data sections
         if "metadata" not in config_store.sections():
             config_store.add_section("metadata")
-        if "next_tokens" not in config_store.sections():
-            config_store.add_section("next_tokens")
+        if "tenant_data" not in config_store.sections():
+            config_store.add_section("tenant_data")
         self._store = config_store
```
markkasaboski: I wonder if we need to validate that these were created and, if not, raise an exception. If we try to set data on one of these sections and it doesn't exist, what happens? Does `add_section()` already do that? 

https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.add_section

Doesn't say. Guaranteed to work I guess. But there is a `has_section()`.

markkasaboski (approved):
Overall, LGTM.

packages/flare/bin/data_store.py
markkasaboski: Another thing might be the magic strings. "metadata", "next_tokens" and "tenant_data" are used in several spots each. Not a showstopper for me.

xvaier: If we try to set on a section that doesn't exist, it will raise an exception, I've tested and this works even with an empty/nonexistent conf file.

I agree for the magic strings, I'll add them to constants.

PR #103, opened 2025-03-18
Add setting for backfilling events on setup

xvaier:
Adds the ability to configure how far in the past we backfill events, without refactoring the fetching logic this will only be configurable the first time the configuration agent is run.

#### With value
Screenshot 2025-03-17 at 9 10 44 PM

#### Default value
Screenshot 2025-03-17 at 9 09 05 PM

### DIFF
--- a/packages/flare/bin/cron_job_ingest_events.py
+++ b/packages/flare/bin/cron_job_ingest_events.py
@@ -1,5 +1,4 @@
 import json
-from optparse import Option
 import os
 import sys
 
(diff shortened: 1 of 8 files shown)


markkasaboski (approved):

packages/flare/bin/cron_job_ingest_events.py
```diff
@@ -142,6 +147,18 @@ def get_api_key(storage_passwords: client.StoragePasswords) -> str:
...
+def get_number_of_days_to_backfill(storage_passwords: client.StoragePasswords) -> int:
+    number_of_days_to_backfill = get_storage_password_value(
+        storage_passwords=storage_passwords,
+        password_key=PasswordKeys.NUMBER_OF_DAYS_TO_BACKFILL.value,
+    )
+
+    try:
+        return int(number_of_days_to_backfill) if number_of_days_to_backfill else 30
+    except:
+        raise Exception("Number of days to backfill not a number")
```
gpratte-flare: We should probably chain the exception:

```py
raise Exception("Number of days to backfill not a number") from e
```

PR #105, opened 2025-10-03
Improve logging and error handling

xvaier:
- Prevents issue with loading event metadata from crashing ingestion.
- Changes the default logging level to improve debugging.
- Attempt to fix a potential memory leak with the ingestion pipeline.

### DIFF
--- a/packages/flare/bin/flare.py
+++ b/packages/flare/bin/flare.py
@@ -78,12 +78,18 @@ class FlareAPI(AuthBase):
             self.logger.debug(event_feed)
             next_token = event_feed["next"]
             for event in event_feed["items"]:
-                if ingest_full_event_data:
-                    event = self._fetch_full_event_from_uid(
-                        uid=event["metadata"]["uid"]
-                    )
-                    time.sleep(1)  # Don't hit rate limit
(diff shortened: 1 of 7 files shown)


packages/flare/bin/flare.py
```diff
@@ -78,12 +78,17 @@ def fetch_feed_events(
             self.logger.debug(event_feed)
             next_token = event_feed["next"]
             for event in event_feed["items"]:
-                if ingest_full_event_data:
-                    event = self._fetch_full_event_from_uid(
-                        uid=event["metadata"]["uid"]
-                    )
-                    time.sleep(1)  # Don't hit rate limit
-                yield (event, next_token)
+                try:
```
gpratte-flare, line 81: We can rewrite as:

```py
                try:
                    if ingest_full_event_data:
                        event = self._fetch_full_event_from_uid(
                            uid=event["metadata"]["uid"]
                        )
                        time.sleep(1)  # Don't hit rate limit
                except:
                    # There is already logging in the _fetch_full_event_from_uid
                    # we want to continue getting the other events even if one fails.
                    pass
                finally:
                    yield (event, next_token)
```

PR #108, opened 2026-05-22
Update splunk version while fixing CI issues

xvaier:
Raising the flareio version so that we use the latest version with tested python 3.9 compatibility.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,8 +17,7 @@ venv: requirements.txt
 	@rm -rf packages/flare/bin/vendor/bin
 	@rm -rf packages/flare/bin/vendor/packaging
 	@rm -rf packages/flare/bin/vendor/*-stubs
-	@rm -rf packages/flare/bin/vendor/charset_normalizer/md.cpython-39-x86_64-linux-gnu.so
-	@rm -rf packages/flare/bin/vendor/charset_normalizer/md__mypyc.cpython-39-x86_64-linux-gnu.so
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -58,8 +58,9 @@ publish: output/flare.tar.gz
 validate: venv-tools
 	@echo "Running Splunk AppInspect..."
 	@echo "If you get an error about \"libmagic\", run \"brew install libmagic\""
-	@venv-tools/bin/splunk-appinspect inspect --ci "output/flare" || \
```
xvaier, line 61: There was an error with a comparaison in this when there was an error in AppInspect.

Makefile
```diff
@@ -17,8 +17,7 @@ venv: requirements.txt
 	@rm -rf packages/flare/bin/vendor/bin
 	@rm -rf packages/flare/bin/vendor/packaging
 	@rm -rf packages/flare/bin/vendor/*-stubs
-	@rm -rf packages/flare/bin/vendor/charset_normalizer/md.cpython-39-x86_64-linux-gnu.so
-	@rm -rf packages/flare/bin/vendor/charset_normalizer/md__mypyc.cpython-39-x86_64-linux-gnu.so
+	@find packages/flare/bin/vendor -type f -name "*x86_64-linux-gnu.so" -delete
```
xvaier, line 20: There might be a better way to do this, but this attempts to fix the previous fix.

Image

aviau (approved):

